PR #78, opened 2026-02-16
Pytorch based Finetuning LLMs (PEFT)

iswaryaalex:
This PR brings in Pytorch Finetuning playbooks that focus of PEFT techniques like SFT full-finetuning, LoRA and QLoRA using pytorch with ROCm backend

- Added 3 scripts in `assets` : train_full_finetuning, train_lora.py and train_qlora.py
- Added `README` to walk through the different techniques, overview, comparison , details, how to finetune, how to customize and finally how to use the finetuned models/adapters
- Note: For Qlora/Lora with bitsandbytes,  the bnb package is not supported for Windows Only works for Linux. 
- Tested on stx halo windows and Linux with ROCm 7.11.0, pytorch
- Added CI/CD tests for playbook README. Note: entire finetuning is not run during testing but covers module imports, initializations etc


adamlam2-amd:
Nice - overall looks decent. Couple thoughts:
1. Why did we choose Gemma3-4B? It's also a gated model so users will need to navigate to HuggingFace to accept Google's licensing. This is not yet clear.
2. We should make it more clear that bitsandbytes isn't supported on Windows
3. The Lora and QLora explanations are a bit weak - I don't think normal users will be able to understand this. If we want to explain it, we can be more thorough. 
4. To compensate for the extra length above, we can save some space on some of the hyperparameter tuning. Instead of showing each value higher or lower, we can just tell them to change it themselves. Essentially, more concise here. 
5. Is the command rocm-smi? I think amd-smi might be the one. 
6. Wandb can be optional since it's a fairly significant addition. 
7. Lastly, if you can add any screenshots anywhere that might be good. Wherever you think is best. Maybe the result of the fine tuning?

playbooks/supplemental/pytorch-finetuning/README.md
```diff
@@ -1,4 +1,376 @@
...
+**Framework**: PyTorch + Hugging Face (Transformers, PEFT, TRL)
+
+> **Note:** You can also try other model architectures, including **GPT-OSS-20B**, by substituting the model in the provided training scripts.
+
+---
+
+## Quick Start
+
+### 1. Install Dependencies
+
```
danielholanda, line 16: Should we recommend the user to use command prompt here?

playbooks/supplemental/pytorch-finetuning/README.md
```diff
@@ -1,4 +1,376 @@
...
+```bash
+pip install transformers==4.57.1 safetensors==0.6.2 datasets==4.2.0 accelerate peft trl "fsspec[http]>=2023.1.0,<=2025.9.0"
+```
+<!-- @test:end -->
+<!-- @os:end -->
+
+### Enable HF Authentication if not using local models
+Authenticate with the Hugging Face Hub to access some model files:
+```bash
+hf auth login
```
danielholanda, line 70: Please add instructions on how to install `hf` as part of the `additional dependencies` section.

playbooks/supplemental/pytorch-finetuning/README.md
```diff
@@ -1,4 +1,376 @@
...
+|-----------------------------------|-------------------|---------------------------------------------------------------------------------------------------------------------|--------------|-------------------------------------------------|
+| [`train_qlora.py`](assets/train_qlora.py)               | **QLoRA** 🌟        | 4-bit quantization + LoRA adapters. Lowest memory use, fastest, small quality trade-off.                            | 12–16GB      | Most users; fast experiments; limited VRAM      |
+| [`train_lora.py`](assets/train_lora.py)                 | **LoRA** 🎯         | Trains small adapter matrices while freezing base model. 3–5x faster; ~95–98% full quality.                         | 24–32GB      | Advanced users; multiple adapters; more VRAM    |
+| [`train_full_finetuning.py`](assets/train_full_finetuning.py) | **Full Fine-tuning** | Updates all model parameters. Maximum quality; highest memory and compute usage.                                    | 40GB+        | Maximum quality; research; large VRAM           |
+
+> **Typical time:** ~15–20 minutes for 800 samples on AMD Strix Halo
+
```
danielholanda, line 151: It’s not immediately obvious to a reader what the model is learning from the dataset or even which dataset is being used.

Adding a short comment/section here would make the script much easier for new readers to understand.

Example:

```
The dataset is converted into chat examples like:

User: “Give me a quote about: <tag>”
Assistant: “<quote> – <author>”

So the fine-tuning is effectively teaching the model to respond to prompts asking for quotes about a topic and return them in the specific format:

`<quote text> - <author>`
```

danielholanda: This will also help the user to have more context before they get to the "customization guide"

playbooks/supplemental/pytorch-finetuning/README.md
```diff
@@ -1,4 +1,376 @@
...
+**2. Reduce Sequence Length:**
+```python
+max_seq_length=256  # Instead of 512
+```
+
+**3. Use More Aggressive Quantization:**
+```
+Full → LoRA → QLoRA
+```
```
danielholanda, line 306: Could you clarify this? The relationship between model quantization and lora techniques is not clear here.

playbooks/supplemental/pytorch-finetuning/playbook.json
```diff
@@ -1,13 +1,12 @@
 {
   "id": "pytorch-finetuning",
-  "title": "Fine-tune with PyTorch",
-  "description": "Fine-tune large language models using PyTorch on STX Halo™",
+  "title": "Fine-tune LLMs with PyTorch and ROCm",
+  "description": "Fine-tune large language models using PyTorch and ROCm on STX Halo™",
   "time": 90,
```
danielholanda: 90min sounds a bit too much for this one. Maybe 45min?
@adamlam2-amd curious on how you feel aboout this one

playbooks/supplemental/pytorch-finetuning/assets/train_full_finetuning.py
```diff
@@ -0,0 +1,194 @@
+#!/usr/bin/env python
+"""
+Full Fine-tuning Gemma 3 4B on AMD Strix Halo (GFX1151)
```
danielholanda, line 3: (No action needed here - just a note for Adam and I)
We will need to make sure this `AMD Strix Halo` is consistent everywhere later

playbooks/supplemental/pytorch-finetuning/README.md
```diff
@@ -1,4 +1,376 @@
...
+|-----------------------------------|-------------------|---------------------------------------------------------------------------------------------------------------------|--------------|-------------------------------------------------|
+| [`train_qlora.py`](assets/train_qlora.py)               | **QLoRA** 🌟        | 4-bit quantization + LoRA adapters. Lowest memory use, fastest, small quality trade-off.                            | 12–16GB      | Most users; fast experiments; limited VRAM      |
+| [`train_lora.py`](assets/train_lora.py)                 | **LoRA** 🎯         | Trains small adapter matrices while freezing base model. 3–5x faster; ~95–98% full quality.                         | 24–32GB      | Advanced users; multiple adapters; more VRAM    |
+| [`train_full_finetuning.py`](assets/train_full_finetuning.py) | **Full Fine-tuning** | Updates all model parameters. Maximum quality; highest memory and compute usage.                                    | 40GB+        | Maximum quality; research; large VRAM           |
```
danielholanda, line 148: To be explored after merge: Is it feasible to add a test that performs a single fine-tuning step?

danielholanda (approved):
Reviewed and tested on Windows! Worked great.

Please wait for Adam's review before merging.

playbooks/supplemental/pytorch-finetuning/README.md
adamlam2-amd: yea I guess we should align on powershell vs command prompt, as they are different.

iswaryaalex:
@danielholanda Adding a hidden test to do single run of lora (model loading, dataset loading and 1 iteration) with a timeout
image
Ran the test locally - looks good

playbooks/supplemental/pytorch-finetuning/README.md
iswaryaalex: This is done 
image

playbooks/supplemental/pytorch-finetuning/README.md
iswaryaalex: Clarified in detail in the description

iswaryaalex:
@adamlam2-amd Thanks for the review. My key changes are

- Added test for single run of LoRa
- Added specific instructions to load gated models like Gemma (we based everything on this for stability.. others were iffy). Install and authenticate HF steps are also added
- Added instructions for "Dataset example" what the format means, how to use it and what we accomplish from finetuning
- Added more details on Lora/Qlora
- Made hyperparam tuning more concise

danielholanda:
@adamlam2-amd Anything else you would like to see here before approving?

playbooks/supplemental/pytorch-finetuning/playbook.json
adamlam2-amd: lets do 45. I'll try it myself also during further testing and have a more accurate time estimate too.

iswaryaalex: I have set this to 45. That is appropriate unless user wants to change params of the training which could extend or shorten

iswaryaalex:
Alright! Merging

PR #149, opened 2026-03-18
Add AMD copyright to files

vgodsoe:
This PR deliberately does not touch the website folder.
JSON files are skipped as well since they're not worth copyrighting.

### DIFF
--- a/playbooks/core/pytorch-rocm-llms/assets/summarizer.py
+++ b/playbooks/core/pytorch-rocm-llms/assets/summarizer.py
@@ -1,3 +1,7 @@
+# Copyright Advanced Micro Devices, Inc.
+# 
+# SPDX-License-Identifier: MIT
+
 """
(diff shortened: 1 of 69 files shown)


playbooks/core/pytorch-rocm-llms/assets/summarizer.py
```diff
@@ -1,3 +1,7 @@
+# Copyright Advanced Micro Devices, Inc.
+# 
+# SPDX-License-Identifier: MIT
```
danielholanda, line 3: Two questions here:
1) Can we make this slightly less verbose by removing the empty commented line everywhere?
2) Is it ok to add the to the bottom of the script instead of the top? Ok if not

vgodsoe: This is the preferred formatting from AMD. I'm just following the guidelines.

PR #167, opened 2026-03-20
Vllm playbook

eddierichter-amd:


playbooks/supplemental/vllm-inference/playbook.json
```diff
@@ -5,16 +5,13 @@
   "time": 90,
   "supported_platforms": {
     "halo": [
-      "linux",
-      "windows"
+      "linux"
```
danielholanda, line 8: I updated this here so only linux instructions show up.

playbooks/supplemental/vllm-inference/README.md
```diff
@@ -28,103 +26,44 @@ This playbook teaches you how to serve LLMs using vLLM on your STX Halo™ GPU a
...
+pip install torch==2.9.1 --extra-index-url https://download.pytorch.org/whl/rocm7.12
 ```
 
-**Verify the download:**
+Install vLLM from the prebuilt ROCm wheel:
 
 ```bash
-ls -lh /data/Qwen3_1_7B
+pip install vllm --extra-index-url https://wheels.vllm.ai/rocm/0.16.1/rocm712
 ```
```
danielholanda, line 64:
```suggestion
### Create a Virtual Environment

Create a Python virtual environment and activate it:

```bash
python -m venv vllm_env
source vllm_env/bin/activate
```

### Installing Basic Dependencies


### Install vLLM

Install vLLM from the prebuilt ROCm wheel:

```bash
pip install vllm --extra-index-url https://wheels.vllm.ai/rocm/0.16.1/rocm712
```
```

Would the following also work?

File: playbooks/supplemental/vllm-inference/README.md [hunk 3]
```diff
@@ -197,13 +115,6 @@ For an interactive chat experience, launch the Gradio client:
 
 Then open your browser to `http://localhost:7860`
```
danielholanda: Is Gradio really the best way to showcase this? If so, please add instructions on how to install it

danielholanda (commented):
Added some initial comments

playbooks/supplemental/vllm-inference/README.md
eddierichter-amd: Yep!

playbooks/supplemental/vllm-inference/README.md
eddierichter-amd: Probably not I was just sticking with what was there. TBH if this is targeted at developers, a script showing how you can communicate with it using the python openai API is probably better. I will update to that.

playbooks/supplemental/vllm-inference/README.md
```diff
@@ -10,160 +10,83 @@ SPDX-License-Identifier: MIT
...
-# Download the model to /data/Qwen3_1_7B
-huggingface-cli download Qwen/Qwen3-1.7B --local-dir /data/Qwen3_1_7B
+python -m venv vllm_env
+source vllm_env/bin/activate
 ```
 
-**Note**: The download may take some time depending on your internet connection. The model is approximately 3.4GB.
-
-**Alternative with custom cache directory:**
+Install PyTorch 2.9.1 with ROCm support:
```
adamlam2-amd: is this the right pytorch version? Halobox is on 2.10

playbooks/supplemental/vllm-inference/README.md
```diff
@@ -10,160 +10,83 @@ SPDX-License-Identifier: MIT
...
-
-Let's use the Qwen3-1.7B model for example:
+Create a Python virtual environment and activate it:
 
 ```bash
-# Install huggingface-cli if not already installed
-pip install -U huggingface_hub
-
-# Download the model to /data/Qwen3_1_7B
-huggingface-cli download Qwen/Qwen3-1.7B --local-dir /data/Qwen3_1_7B
```
adamlam2-amd, line 102: wondering if we can use a newer model, like Qwen 3.5? Thoughts?

playbooks/supplemental/vllm-inference/README.md
```diff
@@ -187,234 +110,55 @@ curl -X POST "http://localhost:8000/v1/chat/completions" \
...
-### Command-Line Options
+Since vLLM exposes an OpenAI-compatible API, you can use the `openai` Python package to interact with it. Install it first:
 
 ```bash
-Options:
-  -p, --port PORT       Port for Gradio UI (default: 7860)
-  -m, --model MODEL     Model name (default: /data/Qwen3_1_7B)
-  -s, --server URL      vLLM server URL (default: http://localhost:8000)
-  -h, --help            Show help message
+pip install openai
```
adamlam2-amd: do we need a venv for this ?

playbooks/supplemental/vllm-inference/README.md
sreeram-11: I have the same question. Any reason for pinning to **PyTorch 2.9.1**? 

Can we change this line to `“Install PyTorch with ROCm support”` ? 

`python -m pip install --index-url https://repo.amd.com/rocm/whl/gfx1151/ torch torchvision torchaudio` installs the latest available version.

playbooks/supplemental/vllm-inference/README.md
sreeram-11: `pip install openai` should be run inside the `vllm_env` virtual environment. 

@eddierichter-amd, can you please either add the venv setup steps or explicitly mention this?

playbooks/supplemental/vllm-inference/README.md
```diff
@@ -10,160 +10,83 @@ SPDX-License-Identifier: MIT
...
 ## Quick Start
 
 ### 1. Start the vLLM Server
 
-Inside the Docker container, start the vLLM server:
+Start the vLLM server:
 
 ```bash
-vllm serve /data/Qwen3_1_7B
+vllm serve Qwen/Qwen3-1.7B
```
sreeram-11, line 88: It looks like the model download steps have been removed. 

Shouldn’t the model be available before running this command? Or is it being downloaded elsewhere?

playbooks/supplemental/vllm-inference/README.md
```diff
@@ -10,160 +10,83 @@ SPDX-License-Identifier: MIT
...
 
 vLLM can be installed in several ways depending on your environment and preferences:
 
-- **Docker (Recommended)** - Use prebuilt container images with ROCm support for AMD GPUs
 - **PyPI Wheel** - Install from Python Package Index using pip
+- **Docker** - Use prebuilt container images with ROCm support for AMD GPUs
 - **Build from Source** - Compile vLLM locally with custom configurations
 
-For this playbook, we'll use the **prebuilt Docker image** which includes vLLM with ROCm support, making it the easiest way to get started on AMD GPUs.
+For this playbook, we'll use the **prebuilt wheel** which includes vLLM with ROCm support, making it the easiest way to get started on AMD GPUs.
```
sreeram-11: Since we already use `“prebuilt container images”` in the **Docker** section, the term `“prebuilt wheel”` here could be confusing. 

It will be clearer to use `“PyPI wheel”` instead so users can easily distinguish between the two.

playbooks/supplemental/vllm-inference/README.md
sreeram-11, line 102: `huggingface-cli` install and model download steps removed. Was this a mistake?

playbooks/supplemental/vllm-inference/README.md
sreeram-11, line 88: I noticed that all instances of `/data/Qwen3_1_7B` were changed to `Qwen/Qwen3-1.7B`. 

Previously, `/data/Qwen3_1_7B` referred to a local directory, but `Qwen/Qwen3-1.7B` looks like a Hugging Face model reference.

1. Do we know where the model is actually downloaded on disk? It would be helpful to call that out for users.
2. My recommendation would be to download the model to a **well-defined local directory** (e.g., `/data/Qwen3_1_7B`, as before) for consistency and clarity. Curious to hear your thoughts on this.

eddierichter-amd: vllm will download the model if it is not there. Is there a reason we would want to explicitly download it first?

playbooks/supplemental/vllm-inference/README.md
eddierichter-amd: It just seemed unnecessary. Should I add this back?

playbooks/supplemental/vllm-inference/README.md
eddierichter-amd: Yeah in all of my using of vllm if it is not installed and the model is a valid hugging face model it will download it to your hugging face repo and then run it, and if it is already downloaded it it will just the downloaded version. Is your experience difference?

playbooks/supplemental/vllm-inference/README.md
eddierichter-amd: My understanding is that the version of vllm is assuming we have torch 2.9.1 installed. If you look at this: https://rocm.docs.amd.com/en/7.12.0-preview/rocm-for-ai/vllm.html?fam=ryzen&gpu=max-pro-395&i=pip&os=linux&os-version=24.04 we specifically install 2.9.1.

playbooks/supplemental/vllm-inference/README.md
eddierichter-amd: Good point! I just updated this, let me know what you think.

playbooks/supplemental/vllm-inference/README.md
eddierichter-amd: For some reason Qwen3.5 wasn't working for me with the wheel. Do you know if it is supposed to work?

playbooks/supplemental/vllm-inference/README.md
eddierichter-amd: We aren't installing from PyPI, right? I am happy to change the word prebuilt to something else.

eddierichter-amd:
Hi @adamlam2-amd and @sreeram-11 thanks for the great comments! I just went through them. Please let me know anymore feedback you have and there are additional changes that I should make.

playbooks/supplemental/vllm-inference/README.md
sreeram-11: > vllm will download the model if it is not there. Is there a reason we would want to explicitly download it first?

If this is the case, it's not necessary.

playbooks/supplemental/vllm-inference/README.md
sreeram-11: I see that you have updated the line to `python -m pip install openai`.

It would be helpful to end users if we can make it explicit that this command should be run inside the already created `.venv`. 
- `python -m pip` will use whichever `python` is currently resolved in the shell, which may not always be the `.venv` interpreter if the environment is not activated.
- Could you please restate the activation step before `python -m pip install openai` 
    - `source .venv/bin/activate` 
    - `python -m pip install openai`
    - Also, add a line above these commands mentioning that `.venv` venv should be activated (if not already) and `openai` should be installed in it

playbooks/supplemental/vllm-inference/README.md
sreeram-11: Okay, I have a couple of documentation clarity suggestions.

1. **Please make the vLLM install method explicit in the “Installing vLLM” section**
    - Right now the section lists:
        - PyPI Wheel
        - Docker
        - Build from Source
    - But the playbook is installing vLLM from **AMD’s ROCm wheel index via pip**. Since that is the method the playbook uses, I think it should be listed explicitly under the installation methods.
    - Under "Installing vLLM", as the first bullet, can you please add: **` AMD ROCm wheel index – Install AMD-provided ROCm-enabled vLLM wheels using pip`**

2. My suggestion for the sentence: `For this playbook, we’ll use the prebuilt AMD ROCm wheel from AMD’s package index, which is the easiest way to get started with ROCm-enabled vLLM on AMD GPUs.`

playbooks/supplemental/vllm-inference/README.md
sreeram-11: 1. Could you please reword the PyTorch/ROCm installation sentence for accuracy?

    - The current line says: `Install ROCm 7.12.0 and PyTorch 2.9.1 in the virtual environment`
    - I think this is misleading, because the command is **not installing the full ROCm 7.12.0 system stack** into the venv. It is installing **PyTorch wheels built for ROCm 7.12.0.**
    - I would recommend changing it to: `Install PyTorch 2.9.1 built for ROCm 7.12.0 in the virtual environment`
    - That wording would make it clearer that:
        - these are Python wheels installed into the venv
        - the command is not itself installing the full system ROCm stack
        - the `+rocm7.12.0` tag refers to the **PyTorch build variant**

2. For clarity, it would help to distinguish the **PyTorch wheel build version** from the **system ROCm installation**, and to encourage users to keep those aligned where possible.

playbooks/supplemental/vllm-inference/README.md
eddierichter-amd: Ooops sorry yeah, I must have forgotten to add that part. Great catch :) Adding it now.

playbooks/supplemental/vllm-inference/README.md
eddierichter-amd: This command also installs the ROCm wheels, right? I am just following this: https://rocm.docs.amd.com/en/7.12.0-preview/rocm-for-ai/vllm.html?fam=ryzen&gpu=max-pro-395&i=pip&os=linux&os-version=24.04 which installs from `https://repo.amd.com/rocm/whl/gfx1151/` which will also install the ROCm wheels. If you see, we set `export PYTHONPATH=.venv/lib/python3.12/site-packages/_rocm_sdk_core/share/amd_smi` 

Sorry I feel like I am being difficult. Happy to change the language but I am pretty sure through the torch wheel we are using the ROCm wheels.

playbooks/supplemental/vllm-inference/README.md
eddierichter-amd: Great changes, Fixed this.

playbooks/supplemental/vllm-inference/README.md
```diff
@@ -10,163 +10,105 @@ SPDX-License-Identifier: MIT
...
+- `--max-num-seqs` - The maximum number of requests vLLM will process at once. vLLM uses continuous batching, meaning it can dynamically add and remove requests from a running batch rather than waiting for an entire batch to finish. This keeps the GPU busy and improves throughput.
 
 ### 2. Test the server with curl
 
-You can test the server using the curl script inside the container:
+You can test the server using the curl script:
 
 ```bash
-./curl_script.sh
+./assets/curl_script.sh
```
sreeram-11, line 111: Should this be `./curl_script.sh` instead? 

I haven't come across other playbooks including the `assets/` prefix when running `bash` commands.

@adamlam2-amd, can you please confirm?

playbooks/supplemental/vllm-inference/README.md
sreeram-11: I see that a new virtual environment (`.openai-venv`) is being created here. Shouldn’t we install `openai` in the existing `.venv` that was created at the beginning of the playbook?

All other dependencies are installed in `.venv`, so it might be better to keep everything consistent and install `openai` there as well.

sreeram-11: 1. So, it should be just these two lines:
```bash
source .venv/bin/activate
python -m pip install openai
```

2. Also, can you update the line above it:
    - **Current version**: `Since vLLM exposes an OpenAI-compatible API, you can use the `openai` Python package to interact with it. Install it first:` 
    - **Recommended**: `Since vLLM exposes an OpenAI-compatible API, you can use the `openai` Python package to interact with it. Activate the `.venv` virtual environment created earlier (if not already active) and install `openai` into it:`

playbooks/supplemental/vllm-inference/README.md
sreeram-11: Thanks, that makes sense, and I think I see the distinction more clearly now.

You are right that this flow is installing not only the `torch` wheel, but also bringing in **ROCm-related Python-distributed components**, which is consistent with the `_rocm_sdk_core` path we later reference through:
`export PYTHONPATH=.venv/lib/python3.12/site-packages/_rocm_sdk_core/share/amd_smi`

So I agree there are **ROCm-related wheels/packages** being installed into the venv.

My concern was mainly about end-user clarity. The current line `Install ROCm 7.12.0 and PyTorch 2.9.1 in the virtual environment` can be read as if the **full ROCm system stack** is being installed into the venv, whereas what is actually happening here is a **pip-based install of PyTorch + ROCm-related Python packages/wheels**.

Would you be okay with slightly refining the wording to make that clearer? 
**My recommendation**: `Install PyTorch 2.9.1 built for ROCm 7.12.0, along with the required ROCm Python packages, in the virtual environment`

playbooks/supplemental/vllm-inference/README.md
eddierichter-amd: Makes sense, just updated it!

playbooks/supplemental/vllm-inference/README.md
eddierichter-amd: Makes sense, just updated it!

playbooks/supplemental/vllm-inference/README.md
eddierichter-amd: I was using assets as that was what was there previously. Happy to update it however folks want.

sreeram-11: Sounds good, let it be as is for now.

Need @danielholanda's confirmation.

PR #195, opened 2026-04-01
Add RCCL clustering playbook for two STX Halo systems

abdmalik-amd:
- Adds a new playbook for clustering two STX Halo systems using RCCL with vLLM
- Covers VRAM allocation (amd-ttm), vLLM installation with ROCm, Ray cluster setup, and serving Qwen3.5-397B-A17B-GPTQ-Int4 across two nodes via tensor parallelism

Note: This playbook is still being tested. Last validated with RCCL via rocm-systems and vLLM 0.17.0+rocm700 wheels, PR marked as draft accordingly


adamlam2-amd:
disregarding this in favour of https://github.com/amd/playbooks/pull/395

PR #229, opened 2026-04-22
Branding Requirements for Launch

vgodsoe:
A couple of missed trademarks that need to be in before we launch.


danielholanda (approved):
Awesome changes. Thanks!

PR #247, opened 2026-04-30
Remove coming soon banners from mock website

danielholanda:
This PR completely removes coming soon banners from the mock website


danielholanda:
@adamlam2-amd Added a quick comment. Contrib is really simple: remove "coming soon" banner from mock website so we can navigate through that more freely.

adamlam2-amd (approved):
thanks - i think this will enable testers and folks to access the halobox specific instructions on the playbook-dev website.

PR #331, opened 2026-06-02
Change default Pytorch model for STX/KRK

danielholanda:
# Description

Change default Pytorch model for STX/KRK.


sreeram-11:
Awesome! Looks great

danielholanda:
@adamlam2-amd 
Can you please also review?
This PR enables STX/KRK with a smaller model and I just want to make sure we are not introducing any bugs to the halo path that should*** be unchanged.

playbooks/core/pytorch-rocm-llms/README.md
```diff
@@ -94,6 +94,15 @@ source llm-env/bin/activate
 
 ### Installing Additional Dependencies
 
+<!-- @device:halo,halo_box -->
+<!-- @var:id=hf_model value="openai/gpt-oss-20b" -->
```
adamlam2-amd: does this install for halo_box? the model is already pre-installed for halo_box, fyi.

adamlam2-amd:
as well, this may be something to think about for the future, but the RAM that comes with the processors is not fixed. For example, a strix halo can come with only 32GB RAM, which would not work with the larger models.

playbooks/core/pytorch-rocm-llms/README.md
adamlam2-amd: if it just sets the var then nevermind haha

adamlam2-amd:
i didn't test it myself; please make sure you do before merging. thanks

danielholanda:
Tested! updating the transformers version was required.

adamlam2-amd:
why was it required? just curious. newer transformer versions might have been breaking other things in windows

PR #344, opened 2026-06-04
Updating the vllm playbook to use prebuilt container

eddierichter-amd:
Updating the vllm playbook to instead use the container


adamlam2-amd:
only for halobox linux right now, we can add non-halobox linux later.

PR #361, opened 2026-06-09
Enable proper synthax highlighting on pws

danielholanda:
# Description

Enable proper synthax highlighting on pws


sreeram-11:
@danielholanda, 

Can we add `c` as well there? Some of the blocks in `pytorch-kernels` playbook use `C`.

Don't we also need `cmd` there?

sdevinenamd:
Looks good!

sdevinenamd (approved):

PR #395, opened 2026-06-15
Update playbook instructions for RCCL clustering with 2 STX Halos

xiaotia1:
Updated instructions based on podman container used for the vLLM playbook. A couple of caveats: the same Podman images being used here use the host network, is running as root, and mount the volume from ~/.local/share/vLLM/models. It has been tested on two Halo boxes.


adamlam2-amd:
should wait on #347 first

adamlam2-amd:
need to adjust it for halobox and include memory and dev center steps

playbooks/supplemental/clustering-rccl/README.md
```diff
@@ -9,6 +9,152 @@ SPDX-License-Identifier: MIT
...
+
+For `amd-ttm` usage examples, see the [ROCm documentation](https://rocm.docs.amd.com/projects/radeon-ryzen/en/docs-7.0.2/docs/install/installryz/native_linux/install-ryzen.html#amd-ttm-usage-examples).
+
+## Locate the network environment variables
+
+> **Note**: Complete this step on both Machine 1 and Machine 2.
+
+On each machine, run `hostname -I | awk '{print $1}'` to find its local IP address.
+
+Now run `ifconfig` and find the name of the network interface that corresponds to the IP address from the previous step and note it down (it will be referred to in the rest of the instructions as `IFNAME`).
```
ldokovic-personal: using `ifconfig` may be sufficient for discovering the netw.interface name. If two Linux machines have the same or very similar PCI/PCIe topology, they will most likely expose the same network interface name, `**enp191s0**`, since each ethernet device is located at PCI bus `191`, slot `0`.

playbooks/supplemental/clustering-rccl/README.md
```diff
@@ -9,6 +9,152 @@ SPDX-License-Identifier: MIT
...
+## Setting network environment variables
+
+> **Note**: Complete this step on both Machine 1 and Machine 2.
+
+These environment variables need to be set to the `IFNAME` prior to starting the Ray cluster and the vLLM server, so that RCCL will use the correct interface for network communication.
+
+```bash
+export NCCL_SOCKET_IFNAME=<IFNAME>
+export GLOO_SOCKET_IFNAME=<IFNAME> 
+```
```
ldokovic-personal, line 142: referring to [#line 77 ](https://github.com/amd/playbooks/pull/395/changes#diff-d19d4caa064baf4c1ee65cbb57bc51b46426429a220d22dbbb252a6ec9e2fa4aR77), do we actually need to query both machines just to obtain the interface name? I agree we still need to query each machine to obtain its IP address, but perhaps we can avoid querying the interface twice and explicitly passing them as ENVS to reduce confusion, since the cluster containers communicate over the same network.

playbooks/supplemental/clustering-rccl/README.md
```diff
@@ -9,6 +9,152 @@ SPDX-License-Identifier: MIT
...
+
+## VLLM container initialization
+
+> **Note**: Complete this step on both Machine 1 and Machine 2.
+
+Podman is a free and open source container tool and your machine has access to a container which has vLLM installed. You can start the container with this command:
+
+```bash
+sudo podman run -it --name vllm_cluster --replace --pull missing --network=host --device /dev/kfd --device /dev/dri -v ~/.local/share/vLLM/models:/opt/vLLM/models --env HF_HOME=/opt/vLLM/models --entrypoint="bin/bash" --shm-size=64g oci-registry.ryai.dev/ryai-vllm:latest
```
ldokovic-personal: In order to preserve the collected data in volume `-v ~/.local/share/vLLM/models:/opt/vLLM/models` and mirror it to /opt/ dir in container,` vLLM/models` folders don’t exist in `~/.local/share` .
so users need to create it manually before pulling the image. We should either choose a directory that already exists for users, or create this path automatically as part of the setup.

playbooks/supplemental/clustering-rccl/README.md
```diff
@@ -9,6 +9,152 @@ SPDX-License-Identifier: MIT
...
+
+At launch, vLLM shards the model across both nodes using tensor parallelism. Once loaded, inference proceeds as if running on a single accelerator.
+
+### Step 1: Start the Ray Head Node (Machine 1)
+
+On Machine 1, start the Ray head node to initialize the cluster:
+
+```bash
+ray start --head --port=6379 --node-ip-address=<MACHINE_1_IP> --num-gpus=1
+```
```
ldokovic-personal, line 160: Ray doesn’t come with vLLM, it's optional dependency [ray] (https://docs.vllm.ai/en/stable/serving/parallelism_scaling/#what-is-ray )
Users might exit the container and try installing Ray separately.
so, instead of installing it via pip 
`pip install “ray[default]” --break-system-packages` - we should stay on level on which podman orchestrate it.

Hence, we should be more explicit to users. They need to stay inside the Podman-managed vLLM container, where Ray is available after the cluster launches. If they exit the container, they should launch it again before running Ray commands, otherwise they may hit _`ray: command not found`_.

playbooks/supplemental/clustering-rccl/README.md
```diff
@@ -9,6 +9,152 @@ SPDX-License-Identifier: MIT
...
+  --max-model-len 32768 \
+  --gpu-memory-utilization 0.9 \
+  --dtype float16 \
+  --tensor-parallel-size 2 \
+  --distributed-executor-backend ray \
+  --enforce-eager \
+  --language-model-only \
+  --reasoning-parser qwen3
+```
+
```
ldokovic-personal, line 191: This is good. But, to wrap up the playbook story, could we also add a concrete validation example for multi-node GPU-to-GPU communication? Specifically, how should users test one node acting as the inference coordinator while a second worker node contributes additional GPU compute/resources

ldokovic-personal (commented):
Well written playbook, clear and concise. A few things we could improve are addressed in this [feedback](https://amdcloud-my.sharepoint.com/:w:/g/personal/ldokovic_amd_com/IQArr6IHKnDERKCyn3YBJT7QAQovbN6PA0BcyVOS2OOor48?e=wl030C&ovuser=3dd8961f-e488-4e60-8e11-a82d994e183d%2Cldokovic%40amd.com&TeamsCID=e39b992b-59ff-4d2f-a214-35765e3ee7e2&clickparams=eyJBcHBOYW1lIjoiVGVhbXMtRGVza3RvcCIsIkFwcFZlcnNpb24iOiI0OS8yNjA1MTQxNjcxNSIsIkhhc0ZlZGVyYXRlZFVzZXIiOmZhbHNlfQ%3D%3D)

playbooks/supplemental/clustering-rccl/README.md
xiaotia1: The concern was that if it is a setup with multiple network interfaces, then selecting the right ethernet interface is not straightforward. How can we point the user to the correct IP and interface in this case?

playbooks/supplemental/clustering-rccl/README.md
xiaotia1: we can query for the IFNAME first and pass them in as environment vars to the podman container, I agree that would streamline it

playbooks/supplemental/clustering-rccl/README.md
xiaotia1: does that mean when the podman image is pulled it does not contain the models folder, and that folder is created with the vllm-launch command? if so, yes I can add it as part of the setup.

playbooks/supplemental/clustering-rccl/README.md
xiaotia1: Noted, will include more info warning users to use Ray within the Podman container.

playbooks/supplemental/clustering-rccl/README.md
ldokovic-personal: yeah, it doesn't
Screenshot 2026-06-18 141447
the one I marked exists inside the container (/opt/vLLM/models) - that's okay. The problem is in this `"~/.local/share/vLLM/models"` path.

I agree, we can add this step as part of the setup, or we can even change the command to point to `~/.local/share/containers` or` ~/.local/share/application.`

playbooks/supplemental/clustering-rccl/README.md
ldokovic-personal: Agree, we can leave it as it is then. It's not a hassle

abdmalik-amd:
@adamlam2-amd @ldokovic-personal, PR is ready for review

PR #420, opened 2026-06-19
Fix RCCL playbook title to match RPC playbook title

abdmalik-amd:
Title seems to have been changed prior to launch, update to match RPC playbook title


danielholanda (approved):
My understanding is that this is a good way of referring to both generic STX Halos as well as our Halo box. @vgodsoe if you have and better suggestions let us know and we can reopen this and change it.

Issue #315, opened 2026-05-28
pip install torch from repo.amd.com/rocm/whl/gfx1151/ fails on Python 3.14


maxdokukin-amd:
### Which playbook is this about?

Running LLMs with Pytorch and AMD ROCm

### Operating system

Windows

### Operating system version

Windows 11 Pro 10.0.26200 Build 26200

### AMD hardware / platform

AMD RYZEN AI MAX+ 395 w/ Radeon 8060S

### GPU driver / ROCm version

Adrenalin 26.5.2

### What happened?

When following the python venv dependencies installation: 
python -m pip install --force-reinstall --no-cache-dir --index-url https://repo.amd.com/rocm/whl/gfx1151/ torch torchvision torchaudio

### What did you expect to happen?

Dependencies installed successfully 

### Steps to reproduce

py -3.14 -m venv llm-env
.\llm-env\Scripts\Activate
python -m pip install --upgrade pip
python -m pip install --force-reinstall --no-cache-dir `
    --index-url https://repo.amd.com/rocm/whl/gfx1151/ `
    torch torchvision torchaudio

### Relevant logs, error messages, or screenshots

```shell
Looking in indexes: https://repo.amd.com/rocm/whl/gfx1151/
ERROR: Could not find a version that satisfies the requirement torch (from versions: none)
ERROR: No matching distribution found for torch
```

### Additional context

The suspected issue cause: python 3.14 is not yet supported. 

Suggested fix: 
When creating python virtual environments, force python version to be 3.12.
Change 
python -m venv llm-env 
to
py -3.12 -m venv llm-env
in 
playbooks/core/pytorch-rocm-llms/README.md
playbooks/core/lmstudio-rocm-llms/README.md

danielholanda:
@maxdokukin-amd Do you have a suggestion on how to properly alert the user what are the required Python versions here?
If so, I would love to see a PR

adamlam2-amd:
HaloBox launches with Py3.13. There is a matching py3.14 for Linux, but I suspect that is because UB 26 comes with Py3.14. Did you just happen to have Py3.14 already installed on your Windows machine? @maxdokukin-amd

maxdokukin-amd:
@adamlam2-amd Strix Halo we have in the lab are with PY3.14. Also the company laptop came with PY3.14

adamlam2-amd:
seems like a rocm issue then that there isnt support for strix halo windows py3.14.

h34v3nzc0dex:
[...]

PR #428, opened 2026-06-22
docs: warn on gfx1151 Windows Python 3.14

Rohithmatham12:
Summary
- Add a visible Windows Python-version note to the shared pytorch dependency for halo/halo_box (gfx1151).
- Add a hidden preflight test before the ROCm PyTorch install that exits with an actionable message on Windows Python 3.14.
- Keep the existing PyTorch package pins and install commands unchanged.

Why
Addresses #315. The current gfx1151 ROCm PyTorch index has Windows wheels for Python cp310-cp313, while Linux includes cp314. On Windows Python 3.14, the playbook currently reaches the pip resolver and fails with a generic no-matching-distribution error. This makes the limitation clear before the install step and tells users to recreate the venv with Python 3.13 or earlier.

Validation
- python3 .github/scripts/validate_playbooks.py
- Extracted pytorch-rocm-llms tests for windows/linux + halo and verified check-gfx1151-python-version is hidden and runs before install-pytorch.
- git diff --check
- Checked https://repo.amd.com/rocm/whl/gfx1151/torch/: torch 2.11.0+rocm7.13.0 Windows tags are cp310-cp313; Linux tags are cp310-cp314.


playbooks/dependencies/pytorch.md
```diff
@@ -8,6 +8,25 @@ SPDX-License-Identifier: MIT
 
 **Install PyTorch with AMD ROCm™ software support** in the created virtual environment:
 <!-- @device:halo,halo_box -->
+> **Windows Python version note:** The current `gfx1151` ROCm PyTorch wheel
```
lucbruni-amd: Please scope the note and test to `halo` only, as `halo_box` pre-installed software includes Python 3.13.

Rohithmatham12:
Thanks for the catch. I updated the branch so the Windows Python 3.14 note and hidden `check-gfx1151-python-version` preflight are scoped to `halo` only.

The shared `install-pytorch` step remains scoped to `halo,halo_box`, so `halo_box` still gets the existing install command without the extra preflight.

Validation run:
- `python3 .github/scripts/validate_playbooks.py`
- extracted `pytorch-rocm-llms` tests for Windows `halo` and `halo_box`; `halo` includes `check-gfx1151-python-version` before `install-pytorch`, while `halo_box` includes only `install-pytorch`
- `git diff --check`

playbooks/dependencies/pytorch.md
sreeram-11: What about `gfx1150` (`stx`) and `gfx1152` (`krk`)? Are Python 3.14 wheels available for these `gfx` targets?

If not, `stx` and `krk` should be added in the `@device` tag along with `halo`.

playbooks/dependencies/pytorch.md
```diff
@@ -7,6 +7,26 @@ SPDX-License-Identifier: MIT
 #### PyTorch
 
 **Install PyTorch with AMD ROCm™ software support** in the created virtual environment:
+<!-- @device:halo -->
+> **Windows Python version note:** The current `gfx1151` ROCm PyTorch wheel
```
sreeram-11: All instructions / tests / notes added in this PR should be scoped to Windows OS (via the `@os` tag).

playbooks/dependencies/pytorch.md
```diff
@@ -7,6 +7,26 @@ SPDX-License-Identifier: MIT
...
+> set includes Windows wheels through Python 3.13. If you are on Windows with
+> Python 3.14, create this virtual environment with Python 3.13 or earlier
```
sreeram-11: If users are running Python 3.14, we should add a command here that shows them how to create a virtual environment using Python 3.13 (or an earlier supported version). 

This information should be visible in the main instructions, so `hidden=True` should not be used.

Rohithmatham12:
Updated to address the review feedback.\n\nChanges in the latest push:\n- Scoped the Python 3.14 warning/preflight to Windows with an @os:windows block.\n- Expanded the affected device scope to halo, stx, and krk after checking the gfx1150/gfx1151/gfx1152 ROCm PyTorch indexes: Windows wheels are available through cp313, while cp314 is Linux-only for the pinned torch 2.11.0+rocm7.13.0 set.\n- Kept halo_box excluded, since that environment ships Python 3.13.\n- Removed hidden=True so the version check is visible.\n- Added the visible Windows command for creating the venv with Python 3.13: py -3.13 -m venv .venv.\n\nValidation run:\n- python3 .github/scripts/validate_playbooks.py\n- Parser check confirmed check-rocm-pytorch-python-version appears for Windows halo/stx/krk, does not appear for Linux, and does not appear for halo_box.\n- git diff --check

playbooks/dependencies/pytorch.md
```diff
@@ -7,6 +7,33 @@ SPDX-License-Identifier: MIT
...
 
 **Install PyTorch with AMD ROCm™ software support** in the created virtual environment:
+<!-- @os:windows -->
+<!-- @device:halo,stx,krk -->
+> **Windows Python version note:** The current `gfx1150`, `gfx1151`, and
+> `gfx1152` ROCm PyTorch wheel sets include Windows wheels through Python 3.13.
+> If you are on Windows with Python 3.14, create this virtual environment with
+> Python 3.13 or earlier before installing PyTorch:
+
+```cmd
```
sreeram-11:
```suggestion

```

playbooks/dependencies/pytorch.md
```diff
@@ -7,6 +7,33 @@ SPDX-License-Identifier: MIT
...
+py -3.13 -m venv .venv
```
sreeram-11:
```suggestion

```

playbooks/dependencies/pytorch.md
```diff
@@ -7,6 +7,33 @@ SPDX-License-Identifier: MIT
...
+.venv\Scripts\activate.bat
```
sreeram-11:
```suggestion

```

playbooks/dependencies/pytorch.md
```diff
@@ -7,6 +7,33 @@ SPDX-License-Identifier: MIT
...
+```
```
sreeram-11: Removing these three lines since I realized that the virtual environment names differ across playbooks.

playbooks/dependencies/pytorch.md
```diff
@@ -7,6 +7,33 @@ SPDX-License-Identifier: MIT
...
+
+<!-- @test:id=check-rocm-pytorch-python-version timeout=30 setup=activate-venv -->
```
sreeram-11:
```suggestion
<!-- @test:id=check-rocm-pytorch-python-version timeout=30 hidden=True setup=activate-venv -->
```

playbooks/dependencies/pytorch.md
```diff
@@ -7,6 +7,33 @@ SPDX-License-Identifier: MIT
 #### PyTorch
 
 **Install PyTorch with AMD ROCm™ software support** in the created virtual environment:
+<!-- @os:windows -->
+<!-- @device:halo,stx,krk -->
+> **Windows Python version note:** The current `gfx1150`, `gfx1151`, and
+> `gfx1152` ROCm PyTorch wheel sets include Windows wheels through Python 3.13.
+> If you are on Windows with Python 3.14, create this virtual environment with
```
sreeram-11:
```suggestion
> If you are on Windows with Python 3.14, create a virtual environment with
```

playbooks/dependencies/pytorch.md
```diff
@@ -7,6 +7,33 @@ SPDX-License-Identifier: MIT
...
+> Python 3.13 or earlier before installing PyTorch:
```
sreeram-11: @danielholanda,

Adding these instructions here could be confusing for end users.

In most playbooks, the first step is to create a virtual environment, followed by the PyTorch dependency installation step, where PyTorch is installed into that environment.

If we add these instructions here at the PyTorch installation step, users running `Python 3.14` would be prompted to create a second virtual environment before installing PyTorch. This creates unnecessary complexity and may lead users to believe that two virtual environments are required.

I think we have two cleaner options:
1. Add this as a note alongside the virtual environment creation step in all playbooks.
2. Update the virtual environment creation steps for STX Halo, STX Point, and Kraken Point to use Python 3.13 by default.

Please let me know your thoughts.

Rohithmatham12:
Updated the branch for the latest review feedback:

- Removed the generic `.venv` creation/activation command block from the PyTorch install dependency section.
- Reworded the note so it does not imply creating a second virtual environment at the PyTorch install step.
- Restored `hidden=True` on `check-rocm-pytorch-python-version`.

Validation:

- `python3 .github/scripts/validate_playbooks.py`
- `git diff --check`
- Extractor check confirmed `check-rocm-pytorch-python-version` is hidden for Windows `halo`/`stx`/`krk`, absent for Linux, and absent for `halo_box`.

sreeram-11:
@danielholanda to approve.

playbooks/dependencies/pytorch.md
danielholanda: Deferring to @adamlam2-amd  here as our lead UX wizard.

adamlam2-amd:
Hi @Rohithmatham12 , thanks for bringing up this issue and for your contribution and work so far. 

Positively, the root cause of this issue (Python 3.14 wheels for Windows not existing), will be internally fixed shortly, and so this warning will not be needed anymore.

Rohithmatham12:
Thanks for the update. Since the underlying Windows Python 3.14 wheel gap is expected to be fixed internally shortly, I agree this warning/preflight should not land and risk becoming stale. I’ll close this PR to keep the queue clean.

PR #435, opened 2026-06-25
Add initial CODEOWNERS file

danielholanda:


harkgill-amd (commented):
Do we want to add codeowners for the other folders in the repos, whoever that may be?

Without that in place, there won't be any reviewers set by default if a PR only touches files in those folders.

danielholanda:
> Do we want to add codeowners for the other folders in the repos, whoever that may be?
> 
> Without that in place, there won't be any reviewers set by default if a PR only touches files in those folders.

Done!

harkgill-amd (approved):
LGTM!

PR #463, opened 2026-06-30
add -y to lms load command to avoid interactive model selection

saman-amd:


sreeram-11:
@saman-amd,

Can you clarify which **model load** command triggered the interactive model selection?

Also, were you trying to load the model **before it had been downloaded on the machine**? In other words, did the interactive model selection appear because the model was not already present locally when you attempted to load it?

I’m trying to understand whether this behavior happens only when the model has not been downloaded yet, or if it can also occur even when the model is already available on the machine.

saman-amd:
> @saman-amd,
> 
> Can you clarify which **model load** command triggered the interactive model selection?
> 
> Also, were you trying to load the model **before it had been downloaded on the machine**? In other words, did the interactive model selection appear because the model was not already present locally when you attempted to load it?
> 
> I’m trying to understand whether this behavior happens only when the model has not been downloaded yet, or if it can also occur even when the model is already available on the machine.

The issue happens if you do not specify the full model name and you have two variants of the model already available on your system, so I think the case was we had gpt-oss 20b and 120b across different playbooks and this was causing the ci systems to be waiting forever for user's input to choose the correct model, the -y will choose the first option
we can make sure that we provide the full model name as well to avoid this
```
lms load qwen

! Multiple models match the provided model key. Please select one.
? Select a model to load |
❯ qwen/qwen3-coder-30b (17.35 GiB)
  qwen/qwen3.5-4b (3.15 GiB)
lms load qwen -y
W 2 models match the provided model key on the same device. Loading the first one.
W Load cancelled.
```

sreeram-11:
@saman-amd,

But in the LM Studio playbook, we've already specified the exact model names:
```
<!-- @var:id=lms_model device=halo,halo_box value="gpt-oss-120b" -->
<!-- @var:id=lms_model device=stx,krk,rx7900xt,rx9070xt,r9700 value="qwen3.5-9b" -->
```

That said, I can see a scenario where it would be useful. If a user has multiple variants of the same model downloaded (for example, **gpt-oss-120b** from `Unsloth`, `ggml-org`, `lmstudio-community`, and `OpenAI`), LM Studio may prompt them to interactively select which checkpoint to use.

PR #464, opened 2026-06-30
Add Driver instructions to Radeon dGPUs

danielholanda:


adamlam2-amd:
would need to add @ require rocm for relevant Radeon devices and workflows that use system ROCm;

playbooks/dependencies/driver.md
sreeram-11, line 1: @danielholanda 

Should we add a link for Adrenalin Software installation (Windows)?

PR #485, opened 2026-07-08
Hermes Firecrawl wrapper 

ldokovic-personal:
A `systemd ` user service that wraps `Firecrawl` and our playbook `Hermes `into a single, manageable Linux OS-level service for HaloBox devices only. 

Includes:

- custom Compose file (`hermes-compose.yaml`) is used as the service entry point
- a clean `.env` template that decouples environment configuration through the Podman. 

The result is a reproducible, easy setup where Firecrawl is configured once and every subsequent start is a single click — or a single command.

This offers seamless integration with our other supplemental playbooks to be registered as services in the same way (such as OpenClaw)


playbooks/supplemental/hermes-lemonade-server/README.md
```diff
@@ -773,9 +773,155 @@ Hermes will now spin up a persistent sandbox container and route all `terminal`
...
+```bash
+[Unit]
+Description=Firecrawl
+After=podman.service
+Requires=podman.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+WorkingDirectory=/home/<your-system-username>/firecrawl
```
sdevinenamd:
```suggestion
WorkingDirectory=${HOME}/firecrawl
```

playbooks/supplemental/hermes-lemonade-server/README.md
```diff
@@ -773,9 +773,155 @@ Hermes will now spin up a persistent sandbox container and route all `terminal`
...
+ExecStart=/usr/bin/podman compose -f hermes-compose.yaml up -d --remove-orphans
+
+# Stop containers when the service stops
+ExecStop=/usr/bin/podman compose -f hermes-compose.yaml down
+
+[Install]
+WantedBy=default.target
+
+```
+>  ⚠️ Replace `your-system-username` with your actual system username.
```
sdevinenamd: Can we avoid these symbols to keep the playbook consistent with others, please?

sdevinenamd: adding ${HOME} at WorkingDirectory=${HOME}/firecrawl will avoid this comment

playbooks/supplemental/hermes-lemonade-server/README.md
```diff
@@ -773,9 +773,155 @@ Hermes will now spin up a persistent sandbox container and route all `terminal`
...
+Make sure the filename matches exactly what you created above, then run:
+```bash
+systemctl --user daemon-reload
+systemctl --user enable firecrawl.service
+```
+If successful, you should see the following output:
+
+> ***Created symlink '/home/<your-system-username>/.config/systemd/user/default.target.wants/firecrawl.service' → '/home/<your-system-username>/.config/systemd/user/firecrawl.service'.***
+
+💡 `default.target.wants/` contains symbolic links to services that are configured to start automatically.
```
sdevinenamd: Can we avoid these symbols to keep the playbook consistent with others, please?

playbooks/supplemental/hermes-lemonade-server/README.md
```diff
@@ -773,9 +773,155 @@ Hermes will now spin up a persistent sandbox container and route all `terminal`
...
+  <img src="assets/podman_health_verification.png" width="500" height="400" />
+</p>
+
+Once verified, bring the stack back down before proceeding:
+```bash
+podman compose -f hermes-compose.yaml down
+```
+Now that everything is validated, start the service through `systemd`:
+```bash
+systemctl --user start firecrawl.service```
```
sdevinenamd:
```suggestion
systemctl --user start firecrawl.service
```

playbooks/supplemental/hermes-lemonade-server/README.md
```diff
@@ -773,9 +773,155 @@ Hermes will now spin up a persistent sandbox container and route all `terminal`
...
+
+[SELF-HOST Firecrawl](https://github.com/firecrawl/firecrawl/blob/main/SELF_HOST.md) is ideal for those who need full control over their scraping and data processing environments but comes with the trade-off of additional maintenance and configuration efforts.
+
+Start by cloning the repository:
+```bash
+git clone https://github.com/firecrawl/firecrawl.git
+```
+> ℹ️ Firecrawl ships with its own `.env` file. We will not be using it. Instead, create your own `.env` file in the root `/firecrawl` directory using the template below.
+
+Create `.env` in the root `/firecrawl` directory:
```
sdevinenamd, line 833: Better to remove all the commented fields. Thoughts?

playbooks/supplemental/hermes-lemonade-server/README.md
```diff
@@ -773,9 +773,155 @@ Hermes will now spin up a persistent sandbox container and route all `terminal`
...
+# ===== System Resource Configuration =====
+# Maximum CPU usage threshold (0.0-1.0). Worker will reject new jobs when CPU usage exceeds this value.
+# Default: 0.8 (80%)
+# MAX_CPU=0.8
+
+# Maximum RAM usage threshold (0.0-1.0). Worker will reject new jobs when memory usage exceeds this value.
+# Default: 0.8 (80%)
+# MAX_RAM=0.8
+```
+As advised:
```
sdevinenamd:
```suggestion
```

playbooks/supplemental/hermes-lemonade-server/README.md
```diff
@@ -773,9 +773,155 @@ Hermes will now spin up a persistent sandbox container and route all `terminal`
...
+> *Protect the admin UI. Set `BULL_AUTH_KEY` to a strong secret, especially on any deployment reachable from untrusted networks.*
```
sdevinenamd:
```suggestion
> Set `BULL_AUTH_KEY` to a strong secret, especially on any deployment reachable from untrusted networks.
```

playbooks/supplemental/hermes-lemonade-server/README.md
```diff
@@ -773,9 +773,155 @@ Hermes will now spin up a persistent sandbox container and route all `terminal`
...
+
+### 3. Deploying Hermes via Compose
+
+Before moving on, make sure you have pulled the latest Hermes Docker image:
+```bash
+podman pull docker.io/nousresearch/hermes-agent:latest
+```
+Once that is done, download the Hermes Compose file [hermes-compose.yaml](assets/hermes-compose.yaml) and place it in the root `/firecrawl` directory:
+
+> ⚠️ This convention is required for systemd to locate and start the service correctly as specified in `WorkingDirectory=/home/<your-system-username>/firecrawl`.
```
sdevinenamd:
```suggestion
> ⚠️ This convention is required for systemd to locate and start the service correctly as specified in `WorkingDirectory=${HOME}/firecrawl`.
```

playbooks/supplemental/hermes-lemonade-server/README.md
```diff
@@ -773,9 +773,155 @@ Hermes will now spin up a persistent sandbox container and route all `terminal`
...
+
+> 💡 You can always expand the stack by adding additional Firecrawl services as needed. The full list of available services can be found in the official [Firecrawl docker-compose.yaml](https://github.com/firecrawl/firecrawl/blob/main/docker-compose.yaml).
+
+### 4. Launch Hermes service through Firecrawl 
+
+#### Health Check
```
sdevinenamd:
```suggestion
```

playbooks/supplemental/hermes-lemonade-server/assets/hermes-compose.yaml
```diff
@@ -0,0 +1,28 @@
+name: firecrawl
+
+services:
+  hermes:
+    image: docker.io/nousresearch/hermes-agent
+    container_name: hermes_agent
+    command: hermes dashboard
```
sdevinenamd:
```suggestion
    command: ["gateway", "run"] 
```

To actually run the gateway. we can start the dashboard separately

playbooks/supplemental/hermes-lemonade-server/README.md
```diff
@@ -773,9 +773,155 @@ Hermes will now spin up a persistent sandbox container and route all `terminal`
 
 ---
 
-## Start Hermes
+<!-- @os:linux -->
+<!-- @device:halo_box -->
+## (Recommended) Hermes Integration with Firecrawl Services
+
+`Firecrawl` is a self-hosted web crawling and content extraction service that unlocks the full potential of Hermes automation. We recommend getting familiar with the [Firecrawl SDK](https://docs.firecrawl.dev/introduction) before diving in.
```
sdevinenamd:
```suggestion
Hermes can browse and extract content from websites using its built-in web tools. However, many modern websites use bot-detection systems, which block simple HTTP requests and return challenge pages instead of the actual content. As a result, Hermes may be unable to reliably extract information from these sites.

To overcome this limitation, `Firecrawl` provides a self-hosted web crawling and content extraction service that can bypass these challenges and unlock the full potential of Hermes automation. We recommend getting familiar with the [Firecrawl SDK](https://docs.firecrawl.dev/introduction) before diving in.
```

How about adding this for motivation to use Firecrawl? 
@superm1

playbooks/supplemental/hermes-lemonade-server/assets/hermes-compose.yaml
```diff
@@ -0,0 +1,28 @@
+name: firecrawl
```
superm1: This seems named wrong.

playbooks/supplemental/hermes-lemonade-server/README.md
superm1: That's good, but I don't think people need to get familiar with the Firecrawl SDK.  I have no idea anything about how the SDK works and I set it up 😛

playbooks/supplemental/hermes-lemonade-server/README.md
```diff
@@ -773,9 +773,155 @@ Hermes will now spin up a persistent sandbox container and route all `terminal`
...
+
+In this setup, Firecrawl runs as a set of Docker containers managed with Podman. To simplify lifecycle management and automatic startup, we register Firecrawl as a user-level `systemd` service that orchetrates the underlying Podman Compose stack. This allows Hermes to start, stop, and verify the Firecrawl service using standard `systemctl --user` commands instead of interacting with containers directly.
```
superm1, line 773: Well I don't think that _Hermes_ needs to start and stop it.  I actually don't think that one user service will have permission to control another user service necessarily.

But it can be configured to start up on it's own by `systemctl enable --user foo.service`.

playbooks/supplemental/hermes-lemonade-server/README.md
superm1: agree

superm1 (commented):
Something I'm wondering - if we're building a compose anyway, shouldn't we just propose a compose that has _both_  firecrawl and hermes in the same container?

sdevinenamd:
> Something I'm wondering - if we're building a compose anyway, shouldn't we just propose a compose that has _both_ firecrawl and hermes in the same container?

As in removing the `(Recommended) Enable Podman Sandboxing` section, since it is already covered in the Firecrawl Compose setup?
I feel we should provide a simple sandboxing workflow[`(Recommended) Enable Podman Sandboxing` ] for users who are just getting started with Hermes. For advanced users, Firecrawl can be presented as the next step.

playbooks/supplemental/hermes-lemonade-server/README.md
ldokovic-personal: totally agree on this

playbooks/supplemental/hermes-lemonade-server/assets/hermes-compose.yaml
ldokovic-personal: I thought users might prefer a GUI over a CLI to land the first time - but I agree that we should start the gateway first and then dashboard. 

I've split Hermes into two separate Compose services - gateway for all actions, and a dashboard so that the user gets the web out-the-box. User does not have to open a separate terminal and execute the "`hermes dashboard`" command because that would negate the benefit of this. Hope that makes sense?

Podman and systemd options are tested and work.

Screenshot 2026-07-09 101525

playbooks/supplemental/hermes-lemonade-server/README.md
ldokovic-personal: Agree! I removed SDK reference and combine these two descriptions

playbooks/supplemental/hermes-lemonade-server/README.md
ldokovic-personal: Yes, did that in https://github.com/amd/playbooks/blob/73812c20aea4e1c21c2f5700640ba79b20723d54/playbooks/supplemental/hermes-lemonade-server/README.md?plain=1#L817

playbooks/supplemental/hermes-lemonade-server/assets/hermes-compose.yaml
ldokovic-personal: yeah, my bad - fixed to `'hermes'`

playbooks/supplemental/hermes-lemonade-server/assets/hermes-compose.yaml
sdevinenamd: Makes sense. Thanks!

sdevinenamd (approved):

PR #486, opened 2026-07-08
Add playbook: Running DeepSeek V4 Flash with ds4

kyuz0:
New playbook for running DeepSeek V4 Flash on Ryzen AI Halo using the ds4 (Dwarf Star 4) inference engine. Includes ds4-cockpit TUI for managing toolboxes, downloading models, and starting servers.

Covers IQ2_XXS imatrix quantization on a single 128GB node.


danielholanda:
Hi @kyuz0. Awesome contribution. I also really liked the YouTube video! We are happy to review and add this to our Playbooks page after some initial testing. Here is our plan:

Step 1 - Our team will review this PR to make sure it has the same look and feel as our other playbooks. We will also test it on device and then merge it.

Step 2 - On a second PR, we will add some tests into this Playbook, so we can ensure this keeps working moving forward. As part of that PR, we can also add some metadata that publishes this on the right platforms.

Once that is done, this Playbook will show up on the official AMD Playbooks website.

kyuz0:
Great! Let me know if you have any questions, I am planning to make a few more of these!

kyuz0:
Thank you for picking this up!

> * This is only for Linux and Strix Halo / HaloBox, right?

Yup, `ds4` is only for Linux, not sure if it'd work for Windows, but definitely the instructions are strictly for Linux. 

> * How many GB should be allocated for GPU (shared, through amd-ttm)? Should it be over 96 (out of 128)?

I set it to 124 in my builds, haven't tested if the default 96 is enough, but it might with the IQ2XXS quant.

> * For the toolbox, it uses ROCm 7.2.4. Any reason for this specific version? Do we have anything on TheRock releases?

I picked the latest stable ROCm for this project. I used TheRock for a long time, but at the moment I am not seeing the need for nightly builds anymore. I could make a TheRock build as well as an option, but I suspect there won't be any performance benefit.

> * Any particular reason for context size = 126,000? Is it the most that could fit on a single node?

TBH it was just a number I typed, I think it can go higher.

> * Is there any particular preference for those ports? (8000, 3000). The reason why I ask is they might be in use already on the HaloBox - so I wonder if they can arbitrarily be changed.

Any port will do for this one.

adamlam2-amd:
Thank you for your reply - will make note of those. 

Happy to merge this, and then we will make one more PR adding auxiliary things (tests, formatting, etc).

playbooks/core/deepseek-v4-flash-ds4/playbook.json
```diff
@@ -0,0 +1,26 @@
...
+    ],
+    "halo": [
+      "linux"
+    ]
+  },
+  "difficulty": "advanced",
+  "isNew": true,
+  "isFeatured": true,
+  "developed": true,
+  "published": true,
```
adamlam2-amd:
```suggestion
  "published": false,
```

Turning this off temporarily while we add the auxiliary playbook items

PR #599, opened 2026-07-20
deps(rocm): restore devel extra in multi-arch rocm.md

saman-amd:
The 7.14 multi-arch migration changed the ROCm install to rocm[libraries,device-gfxNNNN]==7.14.0 and dropped the devel extra. Playbooks that run 'rocm-sdk init' (pytorch-kernels, pytorch-finetuning env-setup) then fail with 'rocm_sdk_devel module ... not installed'. Re-add devel to each device line so the SDK development files are installed, restoring the pre-migration behavior.


adamlam2-amd:
this was merged in another PR, so this one is no longer needed.

sreeram-11:
@saman-amd, 

Thanks for the PR! 

The same changes were added through #591. So, closing this PR.

PR #627, opened 2026-07-24
fix(rccl): add --pids-limit=-1 to podman run command

abdmalik-amd:
- Adds `--pids-limit=-1` to the `podman run` command in the RCCL clustering playbook
- Removes the default container PID cap, which can cause RCCL multi-node workloads to fail mid-inference when the process count exceeds the default limit


sreeram-11:
@adamlam2-amd to approve.

PR #628, opened 2026-07-24
fix(rpc): update VS Build Tools link to VS2022 and rename rpc-server binary

abdmalik-amd:
- Update Visual Studio Build Tools download link to point directly to VS2022 installer for compatibility
- Rename rpc-server to ggml-rpc-server to reflect upstream llama.cpp binary rename


sreeram-11:
@adamlam2-amd to approve.

PR #649, opened 2026-07-28
fix(rpc): remove deprecated -DGGML_HIP_ROCWMMA_FATTN build flag

abdmalik-amd:
- Removes `-DGGML_HIP_ROCWMMA_FATTN=ON` from the llama.cpp cmake command in the RPC clustering playbook
- Removes the corresponding parameter table row, reflecting the flag's upstream deprecation


lucbruni-amd (approved):
Approved based on https://github.com/ggml-org/llama.cpp/pull/26046.

PR #660, opened 2026-08-04
modify header/TOC to be more easily read

adamlam2-amd:
2 changes in this PR:
- manually changes the Kernel playbook headers to promote 'Walkthrough 1' and 'Walkthrough 2' to h2, instead of h3s under the h2 of 'Walkthroughs'. Essentially eliminates one nested level of h3s.

- fairly trivial website change that collapses all h3s until they are being scrolled over. Then they will expand. See the pictures below. 

Justification for this is as follows:
- ~38% shorter TOC, zero navigation lost. At rest only the 268 H2s show; the 168 H3s auto-reveal per section on scroll — open-webui drops 27→14 entries, lemonade 24→10, with every jump target still reachable.
- Keeps the H3s that matter. 37% of H3s are real targets (>50 lines or heavy code, like the two ~500-line kernels walkthroughs); deleting them would force 2+ screens of scrolling in 59% of nested sections.
- No content risk. Pure TOC-component change — no markdown edits, no broken anchors/deep-links, no impact on translations.

image
image


adamlam2-amd:
@harkgill-amd  can you help me merge this? Thanks

PR #661, opened 2026-08-05
Translation: controlling-version disclaimer, and stop translating @github-only blocks

bconsolvo:
Two related changes to what the translation pipeline writes into every localized page.

## 1. Disclaimer now uses controlling-version language

Legal review asked that the disclaimer state the authoritative version explicitly rather than advisorily.

Before:

> **Machine translation.** ... If anything looks wrong, treat the original English playbook as the source of truth.

After:

> **Machine translation.** This page was automatically translated from English and has not been reviewed by a human. It may contain errors, and certain instructions, commands, downloads, product availability, or other content may differ by language or region. In the event of any inconsistency or discrepancy, the original English version of the playbook controls.

- `.github/scripts/disclaimers.json`: new canonical `en` string, regenerated localized string for all **29 locales**, `version` bumped `1` -> `2`.
- **1,073 translated pages** refreshed via `--apply-disclaimers`. No page content was re-translated.

Each localized string uses the standard legal phrasing for "the English version prevails" in that language (French *fait foi et prévaut*, German *ist maßgeblich und hat Vorrang*, Japanese *英語版が優先されるものとします*), keeps `playbook` verbatim as the product term, and keeps the bold lead phrase.

## 2. `@github-only` blocks are no longer translated

The website strips `<!-- @github-only --> ... <!-- @github-only:end -->` outright, so GitHub is the only place these notices render and there is nothing to gain from translating them.

Letting the model rewrite the block had also corrupted it on **41 pages**, which prepended a stray `# ` to the opening marker:

```diff
-# <!-- @github-only -->
+<!-- @github-only -->
```

That is not cosmetic. `stripGitHubOnlyBlocks()` in `website/src/app/api/playbooks/[id]/route.ts` matches from `<!--`, not from the start of the line, so the `# ` survived the strip and rendered as an **empty `<h1>` on the website**.

- `translate_playbook.py` now masks the whole block as a single protected span, so it passes through verbatim.
- New `--sync-github-only` backfill restores the English block (and drops any stray prefix) in translations produced before this rule: **608 files** normalized, no model calls.

## Test plan

- [x] `disclaimers.json` validates: 29 locales, intact `**bold**` markers, `playbook` retained, no untranslated English lead.
- [x] Mask/restore round-trip is byte-identical on all 65 English source files, and the `@github-only` body is no longer present in the translatable text.
- [x] Audit across all translations: 609 `@github-only` blocks, all now byte-identical to their English source, 0 stray prefixes, 0 missing or extra blocks.
- [x] Both `--apply-disclaimers` and `--sync-github-only` are no-ops on re-run (0 files changed).
- [x] Diffs are confined to the disclaimer and `@github-only` lines; the MIT/SPDX header remains verbatim English on every page.
- [x] Human-authored `localized-playbooks/` untouched, as designed.


adamlam2-amd (approved):

bconsolvo:
Thanks for the review @adamlam2-amd. This is now unblocked and green — the issue wasn't your account permissions.

**What happened:** `main` requires branches to be up to date (`strict: true`), so the "Update branch" button was the right instinct. It created merge commit `0ea9fc6` on the fork branch — but **no workflow runs were ever created for that SHA**. Not queued, not the "waiting for approval" kind; the API reported zero runs for it. Since required status checks are evaluated against the PR head commit, the three required contexts (`Playbook Structure & Contract Validation`, `Website Build Check`, `Test Gate (Required Tests)`) had nothing to report and the PR flipped to `BLOCKED`.

The workflows themselves are fine — `Validate Playbooks`, `Check Website`, `Test Playbooks`, and `Check Copyright Notice` all trigger on plain `pull_request` with no path filters and default types that include `synchronize`. Every push to this branch fired all four and passed. It was specifically the server-side "Update branch" merge on a fork PR that generated no CI.

**Fix:** rebased the two commits onto current `main` (dropping the merge commit) and force-pushed. The rebased tree hash is `f490b23e`, identical to the tree of your merge commit, so the content is exactly what you reviewed and approved — nothing added or lost. `main` had only picked up the two OrchestrAI commits (7 files, zero overlap with this PR), so the rebase was clean. Branch protection here has `dismiss_stale_reviews: false` and `require_last_push_approval: false`, so your approval carried over.

All four workflows ran on the new head and passed; the PR is now `MERGEABLE` / `CLEAN` and up to date with `main`.

**For next time:** on a fork PR, if the branch goes stale it's more reliable for the PR author to rebase locally and push than to use the "Update branch" button, since a normal push reliably fires the `synchronize` event that CI depends on.

PR #665, opened 2026-08-05
Make issue templates playbook-scoped and add a playbook proposal form

bconsolvo:
## Summary

Three related changes to how issues work in this repository.

**The bug report was pointing people at ROCm.** It told reporters to file in the ROCm repository and asked for a ROCm version, but the playbooks here span a variety of workflows, so that redirect sent genuine playbook bugs somewhere they would not be fixed. The form is now scoped to playbook content, and asks for the information a maintainer actually needs to reproduce a failure.

**There was no way to pitch a playbook.** A new proposal form lets anyone — community, partners, ISVs, or AMD employees — submit a short abstract and get an AMD decision before writing a full playbook. The goal is to open contributions up while keeping AMD in control of what gets published, and to avoid anyone finishing a playbook we were never going to take.

The ROCm tracker and AMD Community Forum links are replaced by the AMD Developer Discord.

### `bug_report.yml`

Cut from twelve fields to seven, five of which are a dropdown or a single line.

- Dropped the "file this in the ROCm repo instead" redirect and the `rocminfo` / `dpkg | grep rocm` hints.
- The playbook picker is a **generated dropdown** of all 24 `core` + `supplemental` playbooks, instead of free text nobody could filter on.
- Added a required **"which step or section fails?"** field. It was missing, and it is the single most useful thing for reproducing a content bug.
- Hardware and OS are **multi-select dropdowns**, sharing the proposal form's seven device IDs from `playbook.json`. A report can now be routed at the matching runner group in `runners.json` instead of parsing free text. The OS list adds **WSL**, which the proposal form does not have: several playbooks have WSL steps, and folding those reports into "Windows" loses the one detail that says which path failed. A proposal cannot target WSL, since `playbook.json` only supports `windows` and `linux`.
- "What happened", "what did you expect", and "steps to reproduce" are folded into **one** box that prompts for all three.
- Logs and additional context are folded into **one** free-form box at the end, for output, screenshots, and versions. Dropped `render: shell` on it so file attachments and Markdown still work.
- Removed the driver/runtime field and the hardware details field. The latter only existed to describe the "Other AMD hardware" dropdown option, which is gone.

### `playbook_proposal.yml` (new)

Nine fields: title, a three-to-five sentence abstract, what the reader will learn, target devices (including **Other AMD hardware**, with a box to name it), target OSes, difficulty, time, and tags. The intro explains the review flow, states that **the author writes the tests**, and asks people to check open proposals first. The OS field says both Windows and Linux are preferred.

Multi-select **dropdowns** rather than checkboxes for devices and OS on purpose: `required: true` on a checkbox makes that one box mandatory, so checkboxes cannot enforce "pick at least one", while a multi-select dropdown can.

### `gen_issue_template_playbooks.py` (new)

Generates the bug report's playbook dropdown from `playbooks/*/playbook.json`, skipping `backup/` since readers cannot reach drafts. `--check` is wired into `validate-playbooks.yml`, which already runs on every PR, so adding or renaming a playbook without regenerating the list fails CI. Options are emitted as JSON-quoted strings, so a future title containing `: ` cannot break the form.

### `CONTRIBUTING.md` (new)

The full process: bug reports, the abstract-then-PR flow, the three status labels, and what reviewers look for. `playbooks/README.md` previously said we were not accepting proposals and that no new folders should be created, both of which would have blocked an approved contributor at the first step; those are updated.

### Drive-by

Two links in `playbooks/README.md` pointed at `../AMD_BRANDING_GUIDE.md`, which resolves to the repo root. The file is at `playbooks/AMD_BRANDING_GUIDE.md`, so both were 404ing.

## Labels (already done)

Rather than adding a `playbook-proposal` label, the proposal form reuses the existing **`new_playbook`** label ("Pertains to the creation of a new Playbook"), so proposals land in one queue instead of two.

While checking this I found that **`triage` has never existed** as a label in this repository. Both templates asked for it, so GitHub silently dropped it and every bug report filed from the form got only `bug` — across the last 30 issues, `bug` appears 30 times and `triage` zero. The real label was `status: triage`, now renamed to `status::triage` so it matches the `track::` / `os::` / `app::` convention used everywhere else. Renaming carries the label over to existing issues.

Labels are already created on the repository, so nothing needs doing before merge:

| Label | Purpose |
|---|---|
| `new_playbook` | Applied by the proposal form (pre-existing) |
| `status::triage` | Newly filed, not yet reviewed (renamed from `status: triage`) |
| `status::approved` | Go ahead and open a PR linked to the issue |
| `status::needs-info` | Waiting on more detail from the proposer |
| `status::declined` | Not one we will publish |

## Notes for review

- **Issue templates are read from the default branch only**, so none of this is previewable on this branch. The first real test is opening an issue after merge.
- **GitHub cannot enforce field length.** `required` is the only validation the form schema supports, so the proposal form asks for "three to five sentences" as guidance and nothing enforces it.
- The **"Blank issue (Maintainers only)"** entry in the picker is generated by GitHub. `blank_issues_enabled: false` already hides it from the public; maintainers always see it. It cannot be removed.
- I kept the **AMD AI Developer Program** contact link.

### Testing expectation (review follow-up)

`playbooks/TESTING.md` opened with "Playbook creators are not expected to add tests to their playbooks yet," and no file in the repository linked to it. It now states the opposite: **the author of a playbook adds its tests in the same pull request.** That expectation is repeated in the proposal form, in `CONTRIBUTING.md` (with the local `run_playbook_tests.py` command), and in a new short Testing section in the Creation Guide, which links `TESTING.md` for the first time.

This one is worth a second opinion, since it changes what we ask of contributors rather than only how we word it.

## Test plan

- [x] All three templates parse against GitHub's form schema: valid element types, no duplicate ids or options, no reserved `None` option, no unknown attribute keys
- [x] `validate_playbooks.py` passes (34 playbooks, 0 warnings, 0 errors)
- [x] `gen_issue_template_playbooks.py` is idempotent, and `--check` fails with a diff when an option is removed
- [x] `check_copyright.py --require-amd` passes on every new and changed file
- [x] Every relative link in the new and changed Markdown resolves on disk
- [ ] After merge: open a test issue from each template and confirm the picker shows Bug report, Playbook proposal, AMD Developer Discord, and AMD AI Developer Program


.github/archive/ISSUE_TEMPLATE/bug_report.2026-08-05.yml
```diff
@@ -0,0 +1,125 @@
+# Copyright Advanced Micro Devices, Inc.
```
danielholanda: I would recommend not archiving this this way. Simply removing it from the repo and using git to get access to old versions works pretty well.

.github/ISSUE_TEMPLATE/playbook_proposal.yml
```diff
@@ -0,0 +1,145 @@
...
+
+        **How this works**
+
+        1. You submit this abstract.
+        2. AMD playbook maintainers review it and label the issue
+           `status::approved`, `status::needs-info`, or `status::declined`.
+        3. If it is approved, you open a pull request following the
+           [Playbook Creation Guide](https://github.com/amd/playbooks/blob/main/playbooks/README.md)
+           and link it back to this issue. Pull requests without an approved
+           proposal will be closed.
```
danielholanda, line 24: Is there any expectations on who writes the tests?

.github/ISSUE_TEMPLATE/playbook_proposal.yml
```diff
@@ -0,0 +1,145 @@
...
+        in `playbook.json`. Select all that apply.
+      multiple: true
+      options:
+        - halo_box - AMD Ryzen AI Halo Developer Platform
+        - halo - AMD Ryzen AI Max+
+        - stx - AMD Ryzen AI 300 HX
+        - krk - AMD Ryzen AI 300
+        - rx7900xt - AMD Radeon RX 7900 XT
+        - rx9070xt - AMD Radeon RX 9070 XT
+        - r9700 - AMD Radeon AI PRO R9700
```
danielholanda, line 99: You may want to add an "other" option

.github/ISSUE_TEMPLATE/playbook_proposal.yml
```diff
@@ -0,0 +1,145 @@
...
+
+  - type: dropdown
+    id: os
+    attributes:
+      label: Target operating systems
+      description: Select all that apply. Playbooks may cover one or both.
+      multiple: true
+      options:
+        - Windows
+        - Linux
```
danielholanda, line 125: Should likely often be both

danielholanda (approved):
I like the direction here!

.github/archive/ISSUE_TEMPLATE/bug_report.2026-08-05.yml
bconsolvo: Agreed, and done in 3148bccc — `.github/archive/` is gone entirely. The old templates are in the history at 46db044~ if anyone ever wants them.

.github/ISSUE_TEMPLATE/playbook_proposal.yml
bconsolvo: Good catch, there were none, and the repo was actively saying the opposite. `playbooks/TESTING.md` opened with "Playbook creators are not expected to add tests to their playbooks yet," and nothing linked to it from anywhere.

As of 3148bccc the expectation is that **the author writes the tests, in the same pull request as the playbook**, and it now says so in three places a contributor actually passes through:

- this form, in the "How this works" block
- `CONTRIBUTING.md`, in the "Write it" step, with the local `run_playbook_tests.py` command
- the Creation Guide, which now has a short Testing section linking `TESTING.md`

I also flipped that note in `TESTING.md`, since it contradicted the new text head-on. Shout if that is a bigger call than it looks and you would rather keep tests optional for now — it is one paragraph to put back.

.github/ISSUE_TEMPLATE/playbook_proposal.yml
bconsolvo: Added in 3148bccc: **Other AMD hardware (name it below)**, plus an optional free-text field that asks which device it is and whether the proposer has one to test on.

.github/ISSUE_TEMPLATE/playbook_proposal.yml
bconsolvo: Said outright in 3148bccc. The field description now reads: "Preferably both. Most playbooks should run on Windows and Linux, so pick a single OS only when the tooling genuinely does not exist on the other."

Left as a multi-select rather than defaulting both on, since GitHub issue forms cannot preselect dropdown options — but the ask is now explicit rather than implied.

.github/ISSUE_TEMPLATE/playbook_proposal.yml
bconsolvo: Also, as I mentioned, I do intend on putting in a PR to greatly simplify writing CI tests, but I'll address that later.

PR #681, opened 2026-08-07
Add fr-CA (Canadian French) as the 30th playbook locale

bconsolvo:
## Summary

Adds **fr-CA (Canadian French)** as the 30th locale, alongside the existing fr-FR, ahead of the Canada launch. Quebec readers get content in Quebec French rather than France French.

Also fixes a pipeline bug that made any playbook containing an emoji impossible to translate.

## Why the locale name is spelled out

fr-CA and fr-FR are the closest pair in the locale set, and a bare locale code produces something close to a copy of fr-FR. The prompt name is therefore `Canadian French (fr-CA, Quebec/OQLF conventions)`.

The resulting tree is genuinely divergent, not a copy:

| Term | fr-CA | fr-FR |
|---|---|---|
| `clavardage` / `clavarder` (OQLF term for chat) | 54 | 0 |
| `courriel` | 1 | 0 |
| `invite de commande` | 5 | 2 |

Mean text similarity to fr-FR is 0.932. One file (`dependencies/pytorch.md`) is byte-identical, which is correct: it has only 44 words of prose outside code blocks, and none of it differs between the two variants.

## Emoji / non-BMP fix

Two playbooks (`lmstudio-rocm-llms`, `lemonade-getting-started`) contain emoji and failed every attempt with connection resets. Root cause: **the LLM gateway returns HTTP 500 whenever the model's *response* contains a non-BMP character** (U+10000 and above). A request carrying one is fine — only the response fails, which is why retries never helped.

Runs of non-BMP characters are now masked as protected content, like code fences and HTML comments. They are literal UI glyphs rather than translatable prose, so masking costs nothing and additionally guarantees they survive byte-for-byte. Both files then translated cleanly (scores 95 and 90).

This was latent for all 30 locales, not just fr-CA — any new playbook with an emoji would have hit it.

## Verification

- **File parity:** 79 files, exactly matching fr-FR; no files missing or extra.
- **Structural gate:** 0 protected-span mismatches. `@github-only` blocks and license headers are verbatim English, identical to fr-FR's behavior on all four structural checks.
- **Quality:** mean 91.3 (fr-FR 93.1), min 78 (same as fr-FR's min). Two `playbook.json` metadata files sit at 78 and 82 after three remediation passes, where fr-FR has one at 78.
- **No regressions:** re-checked all 30 locales after the shared masking change — 0 protected-span failures, 0 emoji mismatches.
- Masking round-trip is lossless across all 68 source files.

## Note on delivery

This is **content-only**. The website reads exclusively from the canonical English `playbooks/` directory and has no locale selector or routing, so fr-CA is not user-visible until the web team implements locale serving. That work is the critical path for the Canada launch, not translation coverage.


danielholanda (approved):
I can confirm the technique is sound and the modified files are what I expected. That said, I wasn't able to manually verify the translations, as I don't speak French (not the greatest Canadian).

PR #685, opened 2026-08-10
unsloth: fail install-deps on pip error (Windows)

lucbruni-amd:
The Windows `install-deps` block runs two pip installs. If the first (`unsloth[amd] @ git+…`) fails, the second (`triton-windows`) masks it and the block exits 0 — so a failed install surfaces later as a confusing `verify-imports` *'No module named unsloth'* (e.g. #675, #567) instead of a clear install-deps failure.

Fix: `$ErrorActionPreference='Stop'` + explicit `$LASTEXITCODE` checks after each pip so a pip failure fails the step.

Refs #675.

### DIFF
--- a/playbooks/supplemental/unsloth-llms-finetuning/README.md
+++ b/playbooks/supplemental/unsloth-llms-finetuning/README.md
@@ -157,8 +157,11 @@ pip install "unsloth[amd] @ git+https://github.com/unslothai/unsloth.git"
 <!-- @os:windows -->
 <!-- @test:id=install-deps timeout=600 setup=activate-venv -->
 ```powershell
+$ErrorActionPreference = "Stop"
 pip install "unsloth[amd] @ git+https://github.com/unslothai/unsloth.git"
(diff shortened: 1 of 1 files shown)


sreeram-11 (approved):

playbooks/supplemental/unsloth-llms-finetuning/README.md
```diff
@@ -157,8 +157,11 @@ pip install "unsloth[amd] @ git+https://github.com/unslothai/unsloth.git"
 <!-- @os:windows -->
 <!-- @test:id=install-deps timeout=600 setup=activate-venv -->
 ```powershell
+$ErrorActionPreference = "Stop"
```
sreeram-11: These are user facing instructions because `hidden=True` is absent in the `@test` block. We don't want the user to run the three additional lines, do we @lucbruni-amd?

lucbruni-amd: Right. I combined into one command which should correctly expose the failure.

sreeram-11 (approved):

Issue #604, opened 2026-07-20
Consistent, accurate recommended VRAM allocation per device + model


lucbruni-amd:
Recommended VRAM guidance is currently hand-written per playbook, which is one-size-fits-all and open to human error as models and devices vary. We need a way to consistently resolve a recommended VRAM allocation from device + model load and surface it to the user.

Considerations for whatever solution we land on:
- Inputs available from GGUF metadata + context; `overhead` a tunable constant
- CI could print measured VRAM as a drift check to catch calc divergence

lucbruni-amd:
Closing as https://github.com/amd/playbooks/pull/707 landed.

PR #690, opened 2026-08-11
Add validated details and memory requirements

adamlam2-amd:
This PR adds:
- memory requirement for each playbook. Each playbook will have a static minimum memory requirement based on the models used.
- validation details based on manual/automated testing per playbook per device and OS. Details on how that resolves is as follows:

Validated versions — how they resolve

- Built at request time from 3 sources: dependency-versions.json (central version per dep, keyed by device category + OS), registry.json (versionable flag + display name), and each playbook.json's validatedVersions (per-playbook overrides).
- Per device+OS: take the playbook's @require deps (scoped to the matching @os block, versionable only) → look up versions in the central file → apply validatedVersions on top (same-name key overrides, new keys add).
- Net effect: each playbook lists only the software it uses (e.g. Lemonade playbook shows Lemonade, not ROCm). Shared version bumps = one line in the central file.
- Apps show a pinned version on HaloBox and "Latest" on APU/dGPU (via a default fallback); "" hides a row.
Implemented in route.ts (buildValidation); returns resolved validation[device][os].
- Note: override keys must match the registry display name exactly, or they duplicate instead of override.


lucbruni-amd:
Will this close https://github.com/amd/playbooks/issues/604?

adamlam2-amd:
> Will this close #604?

yes

adamlam2-amd:
test

PR #759, opened 2026-08-19
Add a new CI workflow for localized playbooks

hazheng1-amd:
Add localized playbook support to the existing CI workflows for both self-hosted runners and OrchestrAI.

English and localized playbooks share the same workflow and script entry points, while remaining independently selectable, routable, and reportable.

## Self-hosted runners

The existing playbook workflow now:

- Builds a combined test matrix containing English and localized playbooks.
- Supports locale-aware change detection and manual dispatch.
- Routes matrix entries using their configured runner labels.
- Marks localized jobs with `[localized]` in job names.
- Runs both English and localized tests through `.github/scripts/run_playbook_tests.py`.
- Produces locale-aware failure issues and reproduction commands.
- Includes unit tests covering localized playbook selection and execution behavior.

## OrchestrAI

The existing OrchestrAI workflows and scripts now:

- Build and submit both English and localized test batches.
- Pass `PLAYBOOK_LOCALE` and `PLAYBOOK_LOCALIZED_ONLY` to localized jobs.
- Use locale-aware batch IDs and reports.
- Mark localized jobs and PR reports with `[localized]`.
- Support localized PR commands:
  - `!localized-orc`
  - `!localized-orc <locale> all`
  - `!localized-orc <locale> <playbook1,playbook2>`

## Localized playbook behavior

Localized playbooks are stored under:

- `localized-playbooks/<locale>/<category>/<playbook>/`
- `localized-playbooks/<locale>/dependencies/`

By default, a localized playbook uses only dependencies, assets, and files provided by its locale.

Setting the following in the localized `playbook.json` enables fallback to the corresponding English playbook content:

```json
{
  "localized_only": false
}
```

When fallback is enabled, localized metadata and files override their English counterparts, and localized dependencies take precedence over English dependencies.

## Verification content

This PR includes a Simplified Chinese playbook and its localized dependencies for end-to-end CI verification:

- `localized-playbooks/zh-CN/supplemental/pytorch-finetuning/*`
- `localized-playbooks/zh-CN/dependencies/*`


bconsolvo:
The goal here is right, but I'd rather not fork the pipeline. localized-playbooks/ is really just more playbooks to discover — so let's have our existing CI scan playbooks/ and localized-playbooks/<locale>/ and run both through the same mechanism.

Extend `build_test_matrix.py` to also scan the localized tree and emit locale / runner_label / localized_only as matrix fields, pass those into run_playbook_tests.py, and keep `test-playbooks.yml` as the single entry point. The three genuinely-new things (runner label, localized_only inheritance, locale input) are all matrix fields/flags — not a reason for a second workflow.

Why it matters: test-localized-playbooks.yml is ~90% identical to test-playbooks.yml, and the forked build_localized_test_matrix.py swapped our signature-based change detection for filename+regex and dropped the selector-tests job and detection_ok gate — the safety machinery that stops us silently skipping tests. Reusing the canonical scripts gets all of that for free, with one workflow to maintain.

Same goes for the orchestrai_localized_* trio — they mirror the canonical orchestrai_* scripts almost exactly. Let's parameterize those by locale too rather than shipping twins.

hazheng1-amd:
Hi @bconsolvo . I've already integrated the localized workflow into the original CI script in the latest commit. But please note that currently localized CI and native CI will run on the same runners, whether it’s self-hosted or OAI.

adamlam2-amd:
Please resolve conflicts first and ensure CI passes as well.

hazheng1-amd:
@adamlam2-amd Conflicts resolved.

klupink:
Hi, @sreeram-11, we would appreciate it if you could expedite this, as we have downstream work waiting on this. Thanks

bconsolvo:
@adamlam2-amd could you take a look and merge when possible?

adamlam2-amd (approved):

hazheng1-amd:
!localized-orc zh-CN all

hazheng1-amd:
!localized-orc pytorch-finetuning

sreeram-11:
@hazheng1-amd,

Can you please update the PR description to reflect the latest code? The current description seems to be stale.

sreeram-11:
@saman-amd, 

Can you please help review the changes made to OrchestrAI files?

sreeram-11:
@bconsolvo,

Based on the latest commit, the localized jobs require a runner carrying label `localized-zh-cn` (e.g. [self-hosted, Linux, halo, localized-zh-cn] and the Windows equivalent). **Localized** and **native** playbooks share runners, so those `halo` runners must have the extra `localized-zh-cn` label. If they haven't, the required localized jobs queue forever.

FYI, I added the `localized-zh-cn` label to the following hosts:
- `xsj-aimlab-halo-02` (Linux Halo)
- `xsj-aimlab-halo-1` (Windows Halo)

sreeram-11:
LGTM. Approving it!

Four action items before merge:
- [x] **Update the PR description** (@hazheng1-amd)
- [ ] Confirm the **OrchestrAI** changes look good (@saman-amd)
- [x] @lucbruni-amd to resolve my comment
- [x] @harkgill-amd to fix the failing CI tests

sreeram-11 (approved):

.github/workflows/test-playbooks.yml
```diff
@@ -70,9 +75,9 @@ jobs:
         with:
           python-version: '3.13'
 
-      # On a PR, only entries whose extracted @test blocks (or assets) actually
-      # changed are queued; prose-only edits queue nothing. The nightly cron
-      # still runs the full matrix, so a wrongly-skipped entry surfaces in <24h.
+      # Build one matrix containing canonical English and localized entries.
+      # English selection uses extracted test signatures; localized selection
+      # conservatively tracks the effective overlay and dependency trees.
```
sreeram-11, line 80: @lucbruni-amd for visibility.

Can you please confirm if the original logic of running tests only when the `@test` block updates is unchanged?

hazheng1-amd:
@sreeram-11 The PR description has been updated.

sreeram-11:
> Hi, @sreeram-11, we would appreciate it if you could expedite this, as we have downstream work waiting on this. Thanks

@lucbruni-amd, @harkgill-amd, @saman-amd,

Please take a look at [my comment above](https://github.com/amd/playbooks/pull/759#issuecomment-5626077040) and complete the pending action items. Let's merge this ASAP as downstream work is blocked.

.github/workflows/test-playbooks.yml
lucbruni-amd: Yes, unchanged for English tests. For localized selection it seems a typo fix in the README.md will indeed trigger a localized test which matches the wording above.

saman-amd (commented):
Blocking issue for OrchestrAI Path:
When OrchestrAI runs a translated playbook, orchestrai_trigger.py sets an environment variable called PLAYBOOK_LOCALE.
The test library (ucicd-test-library) watches for that variable. When it sees it, it looks for a script called run_localized_playbook_tests.py and runs that instead of the normal one. If the script isn't there, it stops and reports a failure.
This PR doesn't create that script. It took a different approach it added --locale and --localized-only options to the existing run_playbook_tests.py instead. So the file the test library is looking for doesn't exist.
The result: every translated OrchestrAI run acquires a machine, clones the repo, installs dependencies, and then fails with "run_localized_playbook_tests.py is missing." The hardware time is spent before the failure happens.

To fix: either add a small run_localized_playbook_tests.py that forwards to run_playbook_tests.py --locale, or update the test library to call the new script directly.

sreeram-11:
> Blocking issue for OrchestrAI Path:
When OrchestrAI runs a translated playbook, orchestrai_trigger.py sets an environment variable called PLAYBOOK_LOCALE.
The test library (ucicd-test-library) watches for that variable. When it sees it, it looks for a script called run_localized_playbook_tests.py and runs that instead of the normal one. If the script isn't there, it stops and reports a failure.
This PR doesn't create that script. It took a different approach it added --locale and --localized-only options to the existing run_playbook_tests.py instead. So the file the test library is looking for doesn't exist.
The result: every translated OrchestrAI run acquires a machine, clones the repo, installs dependencies, and then fails with "run_localized_playbook_tests.py is missing." The hardware time is spent before the failure happens.
To fix: either add a small run_localized_playbook_tests.py that forwards to run_playbook_tests.py --locale, or update the test library to call the new script directly.

@hazheng1-amd, please resolve this comment by @saman-amd, and then we will be good for merge.

hazheng1-amd:
> Blocking issue for OrchestrAI Path: When OrchestrAI runs a translated playbook, orchestrai_trigger.py sets an environment variable called PLAYBOOK_LOCALE. The test library (ucicd-test-library) watches for that variable. When it sees it, it looks for a script called run_localized_playbook_tests.py and runs that instead of the normal one. If the script isn't there, it stops and reports a failure. This PR doesn't create that script. It took a different approach it added --locale and --localized-only options to the existing run_playbook_tests.py instead. So the file the test library is looking for doesn't exist. The result: every translated OrchestrAI run acquires a machine, clones the repo, installs dependencies, and then fails with "run_localized_playbook_tests.py is missing." The hardware time is spent before the failure happens.
> 
> To fix: either add a small run_localized_playbook_tests.py that forwards to run_playbook_tests.py --locale, or update the test library to call the new script directly.

@saman-amd Hi Saman. We originally added an extra set of scripts to run CI for Localized Playbooks. But now the Localized CI workflow has been merged into the original CI scripts. So please have the test library call `run_playbook_tests.py` directly.

bconsolvo:
@saman-amd — please update ucicd-test-library to call run_playbook_tests.py directly instead of a run_localized_playbook_tests.py. Localized CI now lives in the one script and we'd rather not maintain a second script name.

Where the library looks for run_localized_playbook_tests.py when PLAYBOOK_LOCALE is set, invoke instead:

python .github/scripts/run_playbook_tests.py --playbook "<id>" --platform "<windows|linux>" --device "<device>" --locale "$PLAYBOOK_LOCALE" --localized-only "<true|false>"
Note: the script reads --locale as a flag, not the PLAYBOOK_LOCALE env var, so pass the value through explicitly. An empty --locale selects English, so the same call works for non-localized runs.

Since this fix is in the library, not this PR, the OrchestrAI checks here stay red until it's deployed. What's the timeline — and should we hold #759 or merge and let those checks go green once the library is updated?

bconsolvo:
@saman-amd , we will still close on this PR after tests pass, and just ask you to create another PR when you are back to address the noted issues.

saman-amd:
Let's merge this, I will get the change in test-library

lucbruni-amd:
CI now fully green, would be nice to merge this (226 checks) before other PRs.

bconsolvo:
Thanks for merging, @sreeram-11 .

@saman-amd, please address the latest comments in a new PR when you have a chance.

PR #775, opened 2026-08-20
Add device and runner targeting to Test Playbooks

sreeram-11:
Manually dispatching `test-playbooks.yml` used to be all-or-nothing: you could pick a single playbook, but it always ran across every device (halo, stx, krk, r9700, rx7900xt, rx9070xt) and both OSes the playbook supports. 

There was also no way to pin a run to a specific machine: GitHub just grabbed any free runner from the matching pool.

This PR adds three optional `workflow_dispatch` inputs to fix both gaps.

### New inputs

- **`platform`** — restrict to `linux` or `windows` (default `all`).
- **`device`** — restrict to a single device (default `all`).
- **`runner_label`** — target one specific self-hosted runner by a label unique to it (hostname, IP, or alias).

These compose with the existing `playbook_id`, so you can run e.g. *all playbooks on Linux Halo only*, or *one playbook on one exact machine*. Leaving everything at `all` / empty keeps the current behavior. Nightly cron and PR runs are untouched — these inputs only apply to manual dispatches.

### How runner targeting works

A self-hosted runner is picked only if it carries **every** label in `runs-on`, so appending a label that's unique to one machine pins the job to it. `runner_label` is validated up front against `.github/runners.json`:

- A typo or unknown label **fails fast** in the `detect-changes` job, with the list of valid labels, instead of leaving a job queued for a runner that never appears.
- `runner_label` requires both `platform` and `device` to be set to a specific value (not `all`), otherwise the label can't resolve to a single runner.
- Only labels **unique** to one runner are accepted for targeting. Shared labels like `halo` or `gfx_gfx1201` are intentionally excluded.

To support IP/alias targeting (not just hostname), `runners.json` now carries a `labels` array per runner listing its extra unique labels.

### Visibility

The requested runner label shows up in the job name (e.g. `Test comfyui-image-gen (linux/halo) @xsj-aimlab-halo-02`) and in each job's step summary, so you can always tell what a run targeted without having to remember what you typed.

### Notes

- The IPs/aliases added to `runners.json` were transcribed from the runner settings page. A stale value fails safe (a clear "unknown label" error, never a mis-target).


sreeram-11:
Closing this as it is merged through #786.

PR #794, opened 2026-08-21
ci: skip Qwen 30B on undersized STX systems

saman-amd:
## Summary

- correct the OrchestrAI eligibility comment: `vscode-qwen3-coder` uses LM Studio full Vulkan offload, not Lemonade GPU+CPU offload
- temporarily exclude `vscode-qwen3-coder` from current 32 GB STX systems
- retain the existing model, 32K context, playbook variables, and `--gpu max` policy

This PR changes only `.github/orchestrai-config.yml`. It does not modify either playbook README.

## Failure evidence

In run 32436715530:

- Windows STX failed with `unable to allocate Vulkan0 buffer`; LM Studio exposed 10.71 GiB to Vulkan for an 18.63 GB model before allocating the 32K context.
- Linux STX reached 91%, then both runtime attempts exited before health with `SIGABRT`; kernel diagnostics reported insufficient AMDGPU submission memory.
- The Windows RX 9070 XT LM Studio result never executed and was shared-batch collateral, not a product failure.

Restore STX coverage when the farm provides a system with enough shared memory.

## Validation

- `python3 -m pytest -q .github/scripts/test_build_test_matrix.py`
- 23 passed, 10 subtests passed
- `git diff --check` passed


adamlam2-amd (approved):

adamlam2-amd:
test

PR #797, opened 2026-08-25
update for memory selector and updating memory filters

adamlam2-amd:
### What changed

Some playbooks run a bigger model on Ryzen AI Halo and a smaller one on Strix/Krackan/dGPUs. Two `playbook.json` fields now support this — each is **a plain value OR an object keyed by device**:

- `primaryModel` — a string, or `{ device: model }`
- `recommendedSystemMemory` — a number (GB), or `{ device: number }`

The 18 single-model playbooks are unchanged (still plain values). Three became per-device:

| Playbook | halo / halo_box | stx / krk / dGPUs |
| --- | --- | --- |
| `n8n-automation-gpt-oss` | GPT-OSS-120B · 128 GB | GPT-OSS-20B · 32 GB |
| `lmstudio-rocm-llms` | GPT-OSS-120B · 128 GB | Qwen3.5-9B · 8 GB |
| `pytorch-rocm-llms` | GPT-OSS-20B · 32 GB | Qwen3.5-4B · 8 GB |

### Frontend change needed

The left-column model + memory display must be **device-aware** — it should reflect the device selected in the filter and update when the selection changes.

For each field:

- If it's a **plain value**, show it as-is.
- If it's an **object**, show the entry for the selected device.

```ts
function forDevice<T>(
  field: T | Partial<Record<Device, T>> | undefined,
  device: Device,
): T | undefined {
  if (field == null) return undefined;
  return typeof field === "object"
    ? (field as Record<Device, T>)[device]
    : field;
}
```

Both fields stay omitted on no-model playbooks (`amd-sync`, `cvml`, `pytorch-kernels`).


adamlam2-amd:
Let me first confirm with the web team that a device-aware model/memory display is possible

harkgill-amd (approved):
From a purely backend perspective, these changes LGTM. 

Will leave merging up to @adamlam2-amd as it hinges on his discussion with web team.

PR #819, opened 2026-08-26
gaia-agents: build the advisor agent lazily so an immediate quit exits at once

saman-amd:
## Symptom

`gaia-hardware-advisor-smoke-linux` times out at its 300s ceiling on every Linux device while the first nine gaia tests pass.

```bash
printf 'quit' | python3 hardware_advisor_agent.py > /tmp/gaia_agent_output.txt
```

## Root cause

`main()` constructed `HardwareAdvisorAgent()` **before** the `input()` loop, so the already-buffered `quit` could not be read until construction finished. `HardwareAdvisorAgent.__init__` chains into GAIA's `Agent.__init__`, whose `LemonadeManager` preloads the default model.

## Reproduced on real hardware

Ryzen AI STX, 31 GB, kernel `7.0.0-28-generic`, Lemonade 11.6.0 healthy with **no model loaded**. Environment built exactly as the playbook's shared gaia dependency does (`python3 -m venv` + `pip install amd-gaia`).

| Run | Result |
|---|---|
| **unpatched, cold cache** | `EXIT=124` at **300s** |
| **unpatched, warm** | `EXIT=0` after **591s** |
| **patched** | `EXIT=0` in **685 ms**, `grep -q "Hardware Advisor Agent"` passes |

Last line before the cold-run kill:

```
lemonade_manager._try_preload_with_ctx | lemonade_manager.py:824 |
Preloading 'Gemma-4-E4B-it-GGUF' with ctx_size=32768 on idle Lemonade server
⏳ Loading Gemma-4-E4B-it-GGUF with ctx_size=32768 tokens. This may take a moment (first run downloads the model)...
```

**The 591s number is the important one:** the preload *does* complete, at nearly 2× the 300s ceiling, even with the model already cached. So raising the timeout is not a fix — it would need >10 minutes per device, to load a model the test never uses.

## The change

The agent is constructed on the **first real request** instead of at startup. An interactive user sees the same work, just moved to their first question; `quit` as the first input no longer pays for a model load it never uses. The smoke test then measures startup/shutdown — what it is named for — rather than model download speed.

Confined to `assets/hardware_advisor_agent.py`. No README, model name, context length, or GPU policy touched.

## Second bug found while testing

`input()` raises `EOFError` once piped stdin is exhausted, and the loop's only other handler is a broad `except Exception` that prints and continues — so **an empty stdin spun forever**. Confirmed against the original loop shape in isolation, and on the box:

| `printf "" \| python3 hardware_advisor_agent.py` | |
|---|---|
| before | `EXIT=124` at the 60s bound |
| after | `EXIT=0` in 566 ms |

## Worth a follow-up (needs a README change, so not here)

Both smoke tests assert on `"Hardware Advisor Agent"`, which is printed by the **banner, before initialization**. That assertion cannot distinguish "started and quit cleanly" from "failed to initialize and returned". On Windows the check is `"Hardware Advisor Agent" -or "Agent ready!" -or "Goodbye!"` — so a fast init *failure* would also pass in ~1s. If Windows R9700 is passing in about a second, that is worth confirming rather than assuming it is healthy. Asserting on `Goodbye!` would make both tests meaningful, but the assertions live in the README.

## Validation

```bash
printf 'quit' | python3 hardware_advisor_agent.py    # expect exit 0, sub-second, "Goodbye!"
printf ''     | python3 hardware_advisor_agent.py    # expect exit 0, no spin
```

Narrow rerun: `gaia-agents` on `linux/stx` (and any one other Linux device) — `gaia-hardware-advisor-smoke-linux` should complete in under a second instead of timing out.


saman-amd:
!orc

adamlam2-amd (approved):

PR #820, opened 2026-08-26
Add Collect Runner Specs workflow for full fleet hardware inventory

bconsolvo:
## Summary
Adds a **read-only, on-demand** workflow (`workflow_dispatch`) that inventories the hardware of every runner attached to this repo and reports CPU, memory, and GPU/VRAM per runner — as per-runner artifacts plus a combined step summary.

Motivation: the device tags in `runners.json` (`halo`, `stx`, `krk`, `rx7900xt`, `rx9070xt`, `r9700`) don't record actual specs (CPU model, installed RAM, VRAM). This lets the fleet report its own specs instead of maintaining them by hand or SSHing in.

## What it covers
- **18 device-fleet machines** — via `select_runners.py` (same source of truth as Runner Heartbeat).
- **3 standalone self-hosted boxes** registered to the repo but omitted from `runners.json`: `CS-UCICD-DT208`, `MNB-DGPU-DT3410` (translation), `x-ultra`.
- **`arc-lnx-pub` ARC scale set** — one ephemeral pod, profiling its cgroup CPU/memory limits (clearly labeled as a container, not physical hardware).

## Design notes
- Modeled on `runner-heartbeat.yml`'s fleet-matrix pattern.
- Read-only hardware queries only. The single privileged call is `sudo -n dmidecode` (non-interactive; degrades gracefully if unavailable).
- `fail-fast: false` + `timeout-minutes: 10` + `if: always()` uploads, so one offline runner can't abort or hang the sweep.
- `workflow_dispatch` only — no schedule.

## Why it must merge to run
`workflow_dispatch` workflows are only dispatchable once the file is on the default branch (`main`). After merge: **Actions → Collect Runner Specs → Run workflow**, then read the combined step summary or download the `runner-specs-*` artifacts.

## Test plan
- [ ] Merge to `main`
- [ ] Dispatch and confirm all 4 jobs (`list-runners`, `collect`, `collect-arc`, `combine`) run
- [ ] Verify the combined summary shows specs for the fleet + extras + ARC pod

### DIFF
--- a/.github/workflows/collect-runner-specs.yml
+++ b/.github/workflows/collect-runner-specs.yml
@@ -1,0 +1,146 @@
+# Copyright Advanced Micro Devices, Inc.
+#
+# SPDX-License-Identifier: MIT
+
+name: Collect Runner Specs
+
+# One-off / on-demand hardware inventory of the self-hosted fleet.
+#
+# Fans out over every machine in .github/runners.json (via select_runners.py,
+# the same single-source-of-truth used by Runner Heartbeat) and dumps CPU,
+# memory, and GPU/VRAM info per runner as an artifact. Read-only: it only
+# queries hardware, never changes anything.
+#
+# Run it from the Actions tab (workflow_dispatch). Download the
+# "runner-specs-<name>" artifacts afterwards, or read the combined summary in
+# the "Combine specs" job's step summary.
+
+on:
+  workflow_dispatch:
+
+permissions:
+  actions: write   # upload artifacts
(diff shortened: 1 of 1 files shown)


danielholanda (approved):
Approved, but please see my message offline to confirm that merging this is necessary.

bconsolvo:
Closing — going with the branch-hijack approach instead (repurpose an existing workflow_dispatch workflow on a throwaway branch, no merge needed). Thanks for the tip.

Issue #771, opened 2026-08-20
[Bug]: Playbook: Clustering Two Ryzen™ AI Halos with RCCL can‘t working [TIC-181758]


BroadenX:
### Which playbook is this about?

clustering-rccl - Clustering Two Ryzen™ AI Halos with RCCL

### Which step or section fails?

vLLM will show error

### AMD hardware

halo - AMD Ryzen AI Max+

### Operating system

Linux

### Operating system version

AIOS

### What happened?

Follow Playbook: [Clustering Two Ryzen™ AI Halos with RCCL | AMD AI Playbooks](https://developer.amd.com/playbooks/clustering-rccl/) to set up two node cluster, but failed to load model via vLLM

after running this command:

vllm serve  /opt/vLLM/models/Qwen3.5-397B-A17B-GPTQ-Int4
  --port 7000 \
  --host 0.0.0.0 \
  --max-model-len 32768 \
  --gpu-memory-utilization 0.9 \
  --dtype float16 \
  --tensor-parallel-size 2 \
  --distributed-executor-backend ray \
  --enforce-eager \
  --language-model-only \
  --reasoning-parser qwen3

vLLM will show error, log saved in attachment(already download model file in advance)

https://ontrack.amd.com/plugins/servlet/jeditor_file_provider?imgId=ckupload202608199986194803257909371&fileName=vLLM+error+log+0819.txt

adamlam2-amd:
@abdmalik-amd can you take a quick look?

adamlam2-amd:
Teams are aware and working on this issue. They'll communicate with you soon.

BroadenX:
what's the progress? please give your update

adamlam2-amd:
Teams are working on this but the original author of the playbook is out of office until Friday.

bconsolvo:
The owner of the playbook is Abdullah and he returns to office on Friday. He would be the most help here.

xiaotia1:
Fixed in [Bugfix RCCL cluster stability with Ray backend (#857) · amd/playbooks@70a2e67](https://github.com/amd/playbooks/commit/70a2e67f8d16a57b17437973169f490c624c635f)

PR #857, opened 2026-09-02
Bugfix RCCL cluster stability with Ray backend

xiaotia1:
Addressing https://github.com/amd/playbooks/issues/771, machines will have Ray workers killed when exceeding Ray's memory usage threshold during vLLM engine startup. This will mitigate those situations when the model weights is close to the VRAM capacity and the system RAM is low.
The issue is tracked internally at https://ontrack.amd.com/browse/TIC-181758 and https://ontrack.amd.com/browse/TIC-180544.
Tested on setup with 2 StrixHalo AI boxes, and setup with 2 GorgonHalo boxes.


adamlam2-amd (approved):
Looks like a clean fix - thanks!

Issue #838, opened 2026-08-28
[Bug]: lemonade-chat-linux (0.03s) fail when run Test n8n-automation-gpt-oss (linux/halo) on Gorgon Halo 495W 128G


LYS001256:
### Which playbook is this about?

amd-sync - Remote Development with AMD Sync

### Which step or section fails?

Instell the Dependence and run automatic playbooks

### AMD hardware

halo_box - AMD Ryzen AI Halo Developer Platform

### Operating system

Linux

### Operating system version

AIOS 0814

### What happened?

Running test: lemonade-chat-linux
Language: bash
Timeout: 1200s
============================================================
Working directory: /home/avc/actions-runner/_work/playbooks/playbooks/playbooks/core/n8n-automation-gpt-oss/assets
Command: bash -c...

Code:
           set -euo pipefail
           
           models_json=""
           for i in $(seq 1 120); do
             models_json="$(curl -s --max-time 2 http://127.0.0.1:13305/api/v1/models || true)"
             if [ -n "$models_json" ]; then
               break
             fi
             sleep 1
           done
           
           if [ -z "$models_json" ]; then
             echo "Lemonade server not ready on http://127.0.0.1:13305/"
             exit 1
           fi
           echo "OK: L...

Exit code: 1
Duration: 0.03s
STDOUT (last 10000 chars):
OK: Lemonade server is responding
Model gpt-oss-120b-mxfp-GGUF is not present in Lemonade /api/v1/models.

Test 'lemonade-chat-linux' failed — skipping remaining tests in this playbook.

[test-results-n8n-automation-gpt-oss-linux-halo.zip](https://github.com/user-attachments/files/31544253/test-results-n8n-automation-gpt-oss-linux-halo.zip)

harkgill-amd:
Hey @LYS001256, the failure you're seeing is due to the Gorgon Halo system missing the gpt-oss-120b-mxfp-GGUF model which is a prerequisite to run this playbook. We're updating our CI to better report and manage prereqs, see https://github.com/amd/playbooks/pull/858. You can try the `users/harkgill/prereq-validate` branch out yourself, it'll handle model download and prereq setup itself at runtime.

In the meantime, I've given the `n8n-automation-gpt-oss` playbook a run on a Gorgon Halo system on my end and all tests are passing. Will close this out for now but please let me know if you have any questions or are still running into issues here.

PR #615, opened 2026-07-21
add CI tests for openhands-getting-started playbook

harkgill-amd:
Closes https://github.com/amd/playbooks/issues/601.

Added tests for openhands CI following test patterns from existing playbooks (most notable n8ns and other lemonade based playbooks).

Temporarily forcing `uv` install through CI tests to ensure consistency across all runners.

```
  ┌──────────────────────────┬───────────────────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ Test ID (×linux/windows) │           Step            │                                                                  What it validates                                                                   │
  ├──────────────────────────┼───────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ prereq-clis-*            │ Prerequisites             │ lemonade, node, npm, uv all present                                                                                                                  │
  ├──────────────────────────┼───────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ lemonade-chat-*          │    Verify the Local Model │ Lemonade /api/v1/models responds; model is present + downloaded; deterministic "Reply with exactly: OK" chat (temperature:0, max_tokens:32) succeeds │
  ├──────────────────────────┼───────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ agent-canvas-install-*   │    Install                │ npm install -g @openhands/agent-canvas succeeds (user-owned npm prefix on Linux, no root needed)                                                     │
  ├──────────────────────────┼───────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ agent-canvas-version-*   │    Install                │ agent-canvas CLI is on PATH (--version, falling back to --help)                                                                                      │
  ├──────────────────────────┼───────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ agent-canvas-server-*    │    Launch                 │ Server starts, http://127.0.0.1:8000/ returns HTTP 200, then the process is killed cleanly (frees the shared runner)                                 │
  ├──────────────────────────┼───────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
[...]

PR #858, opened 2026-09-02
feat(ci): validate & self-heal playbook prerequisites on runners before tests

harkgill-amd:
## What

  Adds a pre-flight prerequisite gate that runs on the runner **before** each playbook's tests. For every in-scope `@require` dependency it runs a **validate → (if missing) install → re-validate** loop:

  - **validate passes** → no-op (provisioned runner, hot path)
  - **validate fails + install fixes it** → self-healed, tests proceed
  - **validate fails + can't fix** → fails fast with a *"provisioning issue, NOT a playbook bug"* banner

  ## Why

We're seeing alot of issues when implementing new playbooks (https://github.com/amd/playbooks/pull/615) or testing existing playbooks (https://github.com/amd/playbooks/issues/838)  that relate to test system setup. The current expectation is that the test system already has the dependencies pre-installed but we fail to document this or provide a simple mechanism to ensure this is done.

This simple validation script will only install dependencies in the case where a system is missing them, better handling the current status quo where this case simply fails without indicating missing deps. It'll also help automatically setup all necessary dependencies on fresh runners for our shift to OrchestrAI.

  ## Coverage (uneven by design; extensible)

  - **validate+install:** lemonade, podman, models, nodejs, n8n
  - **validate-only:**  lmstudio (until CI safe install steps determined)
  - **unchecked no-op:** driver, pytorch, comfyui, rocm, etc. — pass through until commands are added to the registry

We can either expand on each dependency (giving it tested `validate` and `install` commands in `playbooks/dependencies/registry.json`) in this PR or in a followup.

### DIFF
--- a/playbooks/core/n8n-automation-gpt-oss/README.md
+++ b/playbooks/core/n8n-automation-gpt-oss/README.md
@@ -53,18 +53,20 @@ n8n includes a **native Lemonade node** (`Lemonade Chat Model`) that provides a 
 <!-- @device:end -->
 
 <!-- @os:windows -->
-<!-- @require:lemonade,nodejs -->
+<!-- @require:lemonade,nodejs,n8n -->
 <!-- @os:end -->
 
 <!-- @os:linux -->
-<!-- @require:lemonade,podman -->
+<!-- @require:lemonade,podman,nodejs,n8n -->
 <!-- @os:end -->
 
 <!-- @device:halo,halo_box -->
+<!-- @require:lemonade-models-gpt-oss-120b -->
 <!-- @var:id=lemonade_model value="gpt-oss-120b-mxfp-GGUF" -->
 <!-- @device:end -->
 
 <!-- @device:stx,krk,rx7900xt,rx9070xt,r9700 -->
+<!-- @require:lemonade-models-gpt-oss-20b -->
 <!-- @var:id=lemonade_model value="gpt-oss-20b-mxfp4-GGUF" -->
 <!-- @device:end -->
 
--- a/playbooks/core/n8n-automation-gpt-oss/README.md
+++ b/playbooks/core/n8n-automation-gpt-oss/README.md
@@ -195,15 +195,10 @@ npm -v
 <!-- @test:end -->
 
 ## Installing n8n
+
+n8n is installed as part of the [prerequisites](#installing-software-prerequisites) above. Confirm it is available:
+
 <!-- @os:windows -->
-Install n8n globally using npm.
-
-> **Note**: You may see some npm warnings. This is expected.
-
-```bash
-npm install -g n8n
-```
-
 <!-- @test:id=n8n-version timeout=60 hidden=True -->
 ```bash
 n8n --version
--- a/playbooks/dependencies/registry.json
+++ b/playbooks/dependencies/registry.json
@@ -53,6 +53,16 @@
         "linux": [],
         "windows": []
       },
+      "validate": {
+        "linux": {
+          "cmd": "lms --help",
+          "expect_rc": 0
+        },
+        "windows": {
+          "cmd": "lms --help",
+          "expect_rc": 0
+        }
+      },
       "versionable": true
     },
     "vllm": {
@@ -90,6 +100,26 @@
           "halo_box"
         ]
       },
+      "validate": {
+        "linux": {
+          "cmd": "export PATH=\"$HOME/.npm-global/bin:$HOME/.local/bin:$PATH\"; n8n --version",
+          "expect_rc": 0
+        },
+        "windows": {
+          "cmd": "n8n --version",
+          "expect_rc": 0
+        }
+      },
+      "install": {
+        "linux": {
+          "cmd": "npm config set prefix \"$HOME/.npm-global\" && npm install -g n8n",
+          "timeout": 1200
+        },
+        "windows": {
+          "cmd": "npm install -g n8n",
+          "timeout": 1200
+        }
+      },
       "versionable": true
     },
     "vscode": {
@@ -124,6 +154,22 @@
           "halo_box"
         ]
       },
+      "validate": {
+        "linux": {
+          "cmd": "curl -sf --max-time 2 http://127.0.0.1:13305/api/v1/health",
+          "expect_rc": 0
+        },
+        "windows": {
+          "cmd": "curl.exe -sf --max-time 2 http://127.0.0.1:13305/api/v1/health",
+          "expect_rc": 0
+        }
+      },
+      "install": {
+        "linux": {
+          "cmd": "sudo add-apt-repository -y ppa:lemonade-team/stable && sudo apt update && sudo apt install -y lemonade-server",
+          "timeout": 900
+        }
+      },
       "versionable": true
     },
     "lemonade-ready": {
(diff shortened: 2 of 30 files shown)


adamlam2-amd (approved):
In theory it makes sense to me - although I didn't review every single change made.

adamlam2-amd:
seems like much of the CI is failing though... do we know why?

harkgill-amd:
> seems like much of the CI is failing though... do we know why?

While the skeleton framework of "look at `@requires`" then validate prereqs is correct, the actual validate and install step needs to be tuned for each dependency. Working through the failing ones here but they'll need to be enabled on a rolling basis.

harkgill-amd:
Seeing unrelated failures such as
```
Exit code: 1
Duration: 0.01s
STDOUT (last 10000 chars):
OK: Lemonade server is responding
Model Qwen3-Coder-30B-A3B-Instruct-GGUF is not present in Lemonade /api/v1/models.
```
on `Test gaia-agents (linux/r9700)`

This is exactly what the PR aims to resolve, will bundle in `@requires` checks and dependency validate + install steps for other playbooks.

lucbruni-amd (approved):
LGTM. In favour of landing this quickly to enable self-healing on our runners which are frequently plagued, and opting to fix-forward any issues that arise.

harkgill-amd:
@sreeram-11, could you please review when you get a chance?

playbooks/core/n8n-automation-gpt-oss/README.md
sreeram-11: Redundant because you added `@require:n8n` in line 56 and 60.

sreeram-11:
> @sreeram-11, could you please review when you get a chance?

@harkgill-amd, thank you for this PR!

As @adamlam2-amd mentioned, based on the PR description this looks good to me. That said, I'll need a bit more time to review the code thoroughly before I can approve.

A couple of scenarios I wanted to raise:
1. Prereqs not declared via `@require`: There are cases where prerequisites aren't always specified through `@require`. They might just be mentioned informally in the playbook instructions or the `README.md`. How would we handle that scenario?

2. **Custom runner images** (e.g., a Halo box as a runner with its own image): If we add a Halo box as a runner with its own image, and there's something wrong with its dependencies, we wouldn't want CI to try to fix it there, right? Based on this PR, it looks like we'd be attempting to modify that custom image, which may not be desirable.

playbooks/core/n8n-automation-gpt-oss/README.md
harkgill-amd: Removed redundant step.

harkgill-amd:
> A couple of scenarios I wanted to raise:
> 
> 1. Prereqs not declared via `@require`: There are cases where prerequisites aren't always specified through `@require`. They might just be mentioned informally in the playbook instructions or the `README.md`. How would we handle that scenario?
> 2. **Custom runner images** (e.g., a Halo box as a runner with its own image): If we add a Halo box as a runner with its own image, and there's something wrong with its dependencies, we wouldn't want CI to try to fix it there, right? Based on this PR, it looks like we'd be attempting to modify that custom image, which may not be desirable.

1. We should try to maintain a standard where all deps are marked with a require. It'll help keep both the CI and user experience consistent across playbooks. If there are cases where this isn't possible, we can look at them specifically and determine how to go forward (maybe branch requires into seperate user facing and CI tags).

2. Yeah, this is something we were running into. My POV is if a dependency is pre-installed on a HaloBox image, we don't try to self-heal it and only validate. We'd be losing out on the ability to self heal that same dep on regular runners. The alternative is adding in gating checks for specific deps not to self heal if a system is deemed as using a custom runner image.

playbooks/core/n8n-automation-gpt-oss/README.md
```diff
@@ -53,18 +53,20 @@ n8n includes a **native Lemonade node** (`Lemonade Chat Model`) that provides a
 <!-- @device:end -->
 
 <!-- @os:windows -->
-<!-- @require:lemonade,nodejs -->
+<!-- @require:lemonade,nodejs,n8n -->
 <!-- @os:end -->
 
 <!-- @os:linux -->
-<!-- @require:lemonade,podman -->
+<!-- @require:lemonade,podman,nodejs,n8n -->
```
sreeram-11, line 60: @adamlam2-amd,

Why was `nodejs` not within `@require` previously for Linux? Can you please confirm if this addition is correct?

playbooks/core/n8n-automation-gpt-oss/README.md
sreeram-11, line 243: This entire content should be moved to `dependencies/n8n.md` for `@require:n8n` above to reflect correctly.

Currently, in `dependencies/n8n.md`, the suggested method to install `n8n` on Linux is `npm install -g n8n` which contradicts the n8n installation through Podman. So, `n8n.md` needs to be updated.

@adamlam2-amd to confirm.

playbooks/dependencies/registry.json
```diff
@@ -90,6 +100,26 @@
...
+          "expect_rc": 0
+        },
+        "windows": {
+          "cmd": "n8n --version",
+          "expect_rc": 0
+        }
+      },
+      "install": {
+        "linux": {
+          "cmd": "npm config set prefix \"$HOME/.npm-global\" && npm install -g n8n",
```
sreeram-11, line 115: Again, seems like `n8n` is installed through Podman on Linux. @adamlam2-amd to confirm.

sreeram-11 (commented):
Would all the content inside these dependencies be reflected on the website if this branch is merged?

Image

We should be careful not to duplicate instructions. For example, I see many model requirement `.md` files (e.g., `lemonade_models_gemma_4_e2b.md`, `lemonade_models_gpt_oss_120b.md`, etc.). If we already have instructions in the main `README.md` of the playbooks for installing these models, then the content of these `.md` files would duplicate those instructions.

If these dependency contents won't be surfaced to the end user, then it's fine. @harkgill-amd to confirm.

playbooks/dependencies/registry.json
```diff
@@ -137,6 +183,543 @@
...
+      ],
+      "file": "lemonade_models_gpt_oss_120b.md",
+      "preinstalled": {
+        "linux": [
+          "halo_box"
+        ],
+        "windows": [
+          "halo_box"
+        ]
+      },
```
sreeram-11, line 202: This block means that the content of `lemonade_models_gpt_oss_120b.md` will be reflected on the website for Halo Box when we use `@require:lemonade_models_gpt_oss_120b.md`. 

Please make sure that these instructions are not duplicated in the existing `README.md` of the `Lemonade` related playbooks.

Same applies for all other dependency files you have added.

playbooks/dependencies/registry.json
```diff
@@ -137,6 +183,543 @@
...
+      ],
+      "file": "openclaw.md",
+      "preinstalled": {
+        "linux": [
+          "halo_box"
+        ],
+        "windows": [
+          "halo_box"
+        ]
+      },
```
sreeram-11, line 708: I don't think OpenClaw comes pre-installed on the Halo Box.

@adamlam2-amd to confirm.

playbooks/dependencies/registry.json
```diff
@@ -137,6 +183,543 @@
...
+      "description": "Fast Python package/environment manager used by Agent Canvas",
+      "category": "runtime",
+      "platforms": [
+        "linux"
+      ],
+      "file": "uv.md",
+      "preinstalled": {
+        "linux": [
+          "halo_box"
+        ],
```
sreeram-11, line 470: TBC

playbooks/dependencies/registry.json
```diff
@@ -137,6 +183,543 @@
...
+      "category": "app",
+      "platforms": [
+        "linux"
+      ],
+      "file": "agent_canvas.md",
+      "preinstalled": {
+        "linux": [
+          "halo_box"
+        ],
+        "windows": []
```
sreeram-11, line 499: Incorrect

playbooks/dependencies/registry.json
```diff
@@ -137,6 +183,543 @@
...
+      ],
+      "file": "ollama.md",
+      "preinstalled": {
+        "linux": [
+          "halo_box"
+        ],
+        "windows": [
+          "halo_box"
+        ]
+      },
```
sreeram-11, line 531: @adamlam2-amd to confirm

playbooks/dependencies/registry.json
```diff
@@ -137,6 +183,543 @@
...
+      ],
+      "file": "ollama_models_gpt_oss_20b.md",
+      "preinstalled": {
+        "linux": [
+          "halo_box"
+        ],
+        "windows": [
+          "halo_box"
+        ]
+      },
```
sreeram-11, line 562: @adamlam2-amd to confirm

playbooks/dependencies/registry.json
```diff
@@ -137,6 +183,543 @@
...
+      ],
+      "file": "hermes.md",
+      "preinstalled": {
+        "linux": [
+          "halo_box"
+        ],
+        "windows": [
+          "halo_box"
+        ]
+      },
```
sreeram-11, line 593: @adamlam2-amd to review

playbooks/dependencies/registry.json
```diff
@@ -137,6 +183,543 @@
...
+      "description": "podman, distrobox, and pipx used to run containerized toolboxes",
+      "category": "framework",
+      "platforms": [
+        "linux"
+      ],
+      "file": "distrobox.md",
+      "preinstalled": {
+        "linux": [
+          "halo_box"
+        ],
```
sreeram-11, line 619: @adamlam2-amd to review

playbooks/dependencies/registry.json
```diff
@@ -137,6 +183,543 @@
...
+      "description": "Container image used by ds4-cockpit to run the ds4 inference engine",
+      "category": "model",
+      "platforms": [
+        "linux"
+      ],
+      "file": "ds4_toolbox_image.md",
+      "preinstalled": {
+        "linux": [
+          "halo_box"
+        ],
```
sreeram-11, line 647: Incorrect.

@adamlam2-amd to review.

playbooks/dependencies/registry.json
```diff
@@ -137,6 +183,543 @@
...
+      "description": "Terminal UI for running DeepSeek V4 Flash toolboxes on Strix Halo",
+      "category": "app",
+      "platforms": [
+        "linux"
+      ],
+      "file": "ds4_cockpit.md",
+      "preinstalled": {
+        "linux": [
+          "halo_box"
+        ],
```
sreeram-11, line 675: Incorrect.

@adamlam2-amd to review.

playbooks/dependencies/registry.json
```diff
@@ -188,6 +771,26 @@
           "halo_box"
         ]
       },
```
sreeram-11, line 773: @adamlam2-amd,

This means that LM Studio does not come pre-installed on the Halo Box but the models do? Is this correct and intended?

playbooks/dependencies/registry.json
```diff
@@ -203,6 +806,45 @@
...
+      ],
+      "file": "lmstudio_models_qwen3_coder_30b.md",
+      "preinstalled": {
+        "linux": [
+          "halo_box"
+        ],
+        "windows": [
+          "halo_box"
+        ]
+      },
```
sreeram-11, line 837: Similar qn here for @adamlam2-amd

playbooks/core/n8n-automation-gpt-oss/README.md
harkgill-amd: On main Linux was `@require:lemonade,podman` and Windows `lemonade,nodejs`. Linux was intended to run n8n containerized via Podman, so host Node wasn't needed. 

 Added nodejs,n8n because the playbook's own Linux test (README:213) runs host `n8n --version`
 
 @adamlam2-amd, to confirm which is correct.

playbooks/core/n8n-automation-gpt-oss/README.md
harkgill-amd: Waiting on Adam's input for n8n specific changes.

harkgill-amd: No longer using `@require`. This is stale

playbooks/dependencies/registry.json
harkgill-amd: Switched to using `@prereq` instead of `@require`. Website rendering is no longer changed by this PR. Each playbook now indicates,
```
<!-- @require = dependency docs rendered on the website; @prereq = CI-only, validated and auto-installed before tests (never rendered) -->
```

playbooks/dependencies/registry.json
harkgill-amd: No longer assuming pre-installed, can fix forward if needed.

playbooks/dependencies/registry.json
harkgill-amd: No longer assuming pre-installed, can fix forward if needed.

playbooks/dependencies/registry.json
harkgill-amd: No longer assuming pre-installed, can fix forward if needed.

playbooks/dependencies/registry.json
harkgill-amd: No longer assuming pre-installed, can fix forward if needed.

playbooks/dependencies/registry.json
harkgill-amd: No longer assuming pre-installed, can fix forward if needed.

playbooks/dependencies/registry.json
harkgill-amd: No longer assuming pre-installed, can fix forward if needed.

playbooks/dependencies/registry.json
harkgill-amd: No longer assuming pre-installed, can fix forward if needed.

playbooks/dependencies/registry.json
harkgill-amd: I've purposefully left it blank as it does not affect the validate + self heal workflow and I'm not sure if this dep is pre-installed or now.

We will need @adamlam2-amd to do a run through of `registry.json` after we update it with commands from Luca's provisioning script.

PR #860, opened 2026-09-02
Update Playbooks (non Reference) to ROCm 10

### LINKED ISSUE #769: Update ROCm/Pytorch install to the next version
when the next version of ROCm comes out, need to update the playbooks

adamlam2-amd:
As ROCm 10 came out on August 26, this PR updates the instructions to use ROCm 10. It will also cause the runners to download ROCm 10 and the linked Pytorch artifacts, causing CI to run and check them too. 

Will fix https://github.com/amd/playbooks/issues/769

### DIFF
--- a/playbooks/dependencies/rocm.md
+++ b/playbooks/dependencies/rocm.md
@@ -22,7 +22,7 @@ sudo reboot
 <!-- @device:halo_box,halo -->
 <!-- @test:id=install-rocm timeout=300 setup=activate-venv -->
 ```bash
-python -m pip install --index-url https://repo.amd.com/rocm/whl-multi-arch/ "rocm[libraries,devel,device-gfx1151]==7.14.0"
+python -m pip install --index-url https://stable.repo.amd.com/rocm/whl-next/ "rocm[libraries,device-gfx1151]==10.0.0"
 ```
 <!-- @test:end -->
 <!-- @device:end -->
@@ -30,7 +30,7 @@ python -m pip install --index-url https://repo.amd.com/rocm/whl-multi-arch/ "roc
 <!-- @device:stx -->
 <!-- @test:id=install-rocm timeout=300 setup=activate-venv -->
 ```bash
-python -m pip install --index-url https://repo.amd.com/rocm/whl-multi-arch/ "rocm[libraries,devel,device-gfx1150]==7.14.0"
+python -m pip install --index-url https://stable.repo.amd.com/rocm/whl-next/ "rocm[libraries,device-gfx1150]==10.0.0"
 ```
 <!-- @test:end -->
 <!-- @device:end -->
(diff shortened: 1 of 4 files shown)


playbooks/dependencies/rocm.md
```diff
@@ -22,41 +22,41 @@ sudo reboot
 <!-- @device:halo_box,halo -->
 <!-- @test:id=install-rocm timeout=300 setup=activate-venv -->
 ```bash
-python -m pip install --index-url https://repo.amd.com/rocm/whl-multi-arch/ "rocm[libraries,devel,device-gfx1151]==7.14.0"
```
harkgill-amd, line 25: I've retriggered the CI to pickup the latest round of runner fixes but the `pytorch-finetuning` and `pytorch-kernels` failures look to be specific to the ROCm 10 changes.

```
  Error:
  rocm-sdk init
  → ERROR running init: rocm_sdk_devel module for the ROCm SDK development package
    is not installed. This can typically be obtained by installing `rocm[devel]`.
```
is coming from the fact that we're dropping the `devel` package with the shift to ROCm 10. Any reason for that or can we add it back in?

harkgill-amd (changes requested):
Some of the CI failures need to be addressed.

playbooks/dependencies/rocm.md
sreeram-11: Unsloth Finetuning failure also looks specific to ROCm 10 changes.

sreeram-11: @harkgill-amd, `pytorch-finetuning` seems to pass.

sreeram-11:
@adamlam2-amd, please reach out to the authors of both `pytorch-kernels` and `unsloth-llms-finetuning` and help fix the error.

adamlam2-amd:
Pytorch kernels pins its own ROCm version, which Satya is addressing and looking to upgrade anyway. 

Unsloth has its own bug where the torch version is capped, and so it will produce errors in the installation. It will ideally be fixed by Unsloth, but I can also add a disclaimer in a separate.

We need to merge this PR asap else we are recommending an outdated/stale version of ROCm which is not the latest.

playbooks/dependencies/rocm.md
adamlam2-amd: devel should only be needed for pytorch-kernels, which should have its own separate pinned install in its instructions.

sreeram-11 (approved):

harkgill-amd (approved):

adamlam2-amd:
Can we force this through please @sreeram-11

sreeram-11:
@adamlam2-amd, I was expecting CI tests of `unsloth-llms-finetuning` playbook to fail across all target devices.

However, it seems like **R9700** (Windows and Linux) and **Windows RX 9070XT** are passing. Any idea why?

PR #868, opened 2026-09-03
docs(clustering-rccl): correct Ray OOM prevention explanation

abdmalik-amd:
Replaces the inline HTML-formatted note with a dedicated subsection.
Clarifies that the issue is specific to the Ryzen AI Halo's unified
memory pool, and improves the explanation of RAY_memory_monitor_refresh_ms=0.


adamlam2-amd (approved):
thanks - lgtm

Issue #921, opened 2026-09-14
[Bug]: AMD Adrenalin SW Screenshots out of date - no memory adjuster


bconsolvo:
### Which playbook is this about?

pytorch-rocm-llms - Running LLMs on PyTorch with AMD ROCm™ Software

### Which step or section fails?

Adrenalin SW steps out of date

### AMD hardware

halo - AMD Ryzen AI Max+

### Operating system

Windows

### Operating system version

-

### What happened?

The Adrenalin SW screenshots all need to be updated with the latest instructions. There is no memory adjuster any more.
Screenshot in the playbook:
<img width="1693" height="893" alt="Image" src="https://github.com/user-attachments/assets/11fc33df-c4ea-443d-a055-e4255d534516" />

Guruprasad MP reported the new Adrenalin SW screenshots:

<img width="2864" height="1270" alt="Image" src="https://github.com/user-attachments/assets/fdec4f81-a1fd-4a97-a8c4-45203630cfd2" />

bconsolvo:
I am working on this issue, and it is across many playbooks.

bconsolvo:
Being fixed in #934 — removes the obsolete AMD Adrenalin Variable Graphics Memory instructions and screenshots (the control has been removed from Adrenalin), and moves memory adjustment to the BIOS (Windows) / `amd-ttm` (Linux).

Issue #706, opened 2026-08-13
[Bug]: Comfy UI Playbook - cannot adjust memory with Adrenalin


bconsolvo:
### Which playbook is this about?

comfyui-image-gen - Generating Images with ComfyUI and Z Image Turbo

### Which step or section fails?

There is no "Tuning" for Adrenalin SW on my laptop

### AMD hardware

halo - AMD Ryzen AI Max+

### Operating system

Windows

### Operating system version

Windows 11 Enterprise Insider Preview (Build 26300.8553)

### What happened?

Here is what my Adrenalin looks like:

<img width="961" height="605" alt="Image" src="https://github.com/user-attachments/assets/ca366f2c-d799-4029-87af-e06d98c395ca" />

I do not have the setting outlined in the playbook here: 
<img width="886" height="560" alt="Image" src="https://github.com/user-attachments/assets/5a355350-fdcf-4b24-a227-1fa257e22da6" />

adamlam2-amd:
Which OEM laptop are you using? Some OEMs block it

bconsolvo:
HP. Others have experienced this issue as well on other Playbooks. The UI is totally different now for Adrenalin.

bconsolvo:
Being fixed in #934 — removes the obsolete AMD Adrenalin Variable Graphics Memory instructions and screenshots (the control has been removed from Adrenalin), and moves memory adjustment to the BIOS (Windows) / `amd-ttm` (Linux).

PR #934, opened 2026-09-15
Remove obsolete Adrenalin VRAM instructions; use BIOS + amd-ttm

bconsolvo:
## Problem
The **Variable Graphics Memory** control has been removed from AMD Software: Adrenalin Edition (confirmed on current hardware — the Performance → Tuning panel no longer has it). Our playbooks still tell users to open *Performance → Tuning → Variable Graphics Memory*, which no longer exists. Two additional real-world findings from testing:
- The Adrenalin installer may not create a **Start menu shortcut** — it has to be launched from `C:\Program Files\AMD\CNext\CNext\RadeonSoftware.exe`.
- Memory adjustment now has to happen in the **BIOS** (Windows) or via **amd-ttm** (Linux), not in Adrenalin.

## Changes
Almost all playbooks share memory-config content through `@require:memory-config`, so the main fix is one file:

- **`playbooks/dependencies/memoryconfig.md`** — reworked into two sections (fixes 19 playbooks at once):
  - **Install AMD Software: Adrenalin Edition** — download link + the RadeonSoftware.exe launch note (Windows).
  - **Advanced: increasing GPU memory (optional)** — positively framed; Windows = BIOS UMA Frame Buffer Size, Linux = BIOS-minimum + `amd-ttm` GTT/TTM pool, with a kernel 6.16.9+ caveat and a "leave ~20% RAM for the OS" note. Includes a note that some OEMs lock the BIOS setting.
- **3 clustering playbooks** — consolidated their inline copies onto `@require:memory-config`, keeping only the cluster-specific target values (96 GB / `amd-ttm --set 120`) and the per-machine note.
- **Retired screenshots** — deleted `adrenalin_vram_new.png` and `linux_mem_new.png`; no replacements.

## CI presence checks
Added two hidden, non-failing `@test` blocks to the install section so CI confirms the runner is provisioned (they inherit into all `@require:memory-config` playbooks):
- **Windows:** `Test-Path` on `RadeonSoftware.exe` at the documented path.
- **Linux:** `rocminfo | grep gfx` confirms the GPU is reachable (there is no Adrenalin on Linux).

Both are `hidden=True` and `continue_on_error=true`, so a runner missing the software reports the gap without hard-blocking unrelated playbooks. Verified with the CI extractor that each runs only on its own OS and is discovered through `@require` inlining (same mechanism `driver.md` uses).

## Verification
Confirmed in the website dev server: Windows/Linux gating shows the correct steps per OS, no directive comments leak, no references to the deleted screenshots.

## Open question for reviewers
Windows advanced path uses **BIOS UMA Frame Buffer Size**. Some OEM images (e.g. Lenovo) may lock this — a note is included, but flagging since it ties into the Lenovo device triage.


bconsolvo:
Closes #921 — retires the out-of-date Adrenalin memory-adjuster screenshots and the "Performance → Tuning → Variable Graphics Memory" instructions (that panel no longer exists).

Closes #706 — the "no Tuning tab / cannot adjust memory with Adrenalin" report is resolved by moving memory adjustment to the BIOS (Windows) / `amd-ttm` (Linux), which no longer depends on the removed Adrenalin control.

adamlam2-amd:
I think this is slightly incorrect - what machine are you using? The OEM 'pro' products such as HP will hide this tab, but most other systems should have the tab that allows for changing carveout within Adrenalin.

bconsolvo:
> I think this is slightly incorrect - what machine are you using? The OEM 'pro' products such as HP will hide this tab, but most other systems should have the tab that allows for changing carveout within Adrenalin.

The problem is that I have a number of folks telling me that they can't adjust this setting. I don't know which systems it does and doesn't work on. But the way that I described it in this PR should work on any Windows system with the BIOS. What would you suggest, @adamlam2-amd ?

adamlam2-amd:
> > I think this is slightly incorrect - what machine are you using? The OEM 'pro' products such as HP will hide this tab, but most other systems should have the tab that allows for changing carveout within Adrenalin.
> 
> The problem is that I have a number of folks telling me that they can't adjust this setting. I don't know which systems it does and doesn't work on. But the way that I described it in this PR should work on any Windows system with the BIOS. What would you suggest, @adamlam2-amd ?

We can mention the BIOS instructions - I recommend having both there, with Adrenalin as the main option, and BIOS as a backup if Adrenalin doesn't have it.

The far majority of systems should have the Adrenalin setting present.

sreeram-11:
@harkgill-amd, can you please take a look at the failing CI tests here?

bconsolvo:
@adamlam2-amd an you provide an updated screenshot to me with the latest Adrenalin version that you are seeing with this setting, so I can add it to this PR? I don't see it.

adamlam2-amd:
What is wrong with the current screenshot?

bconsolvo:
@adamlam2-amd : I have never seen this on any device where I've installed Adrenalin. I just wanted you to make sure that you are actually seeing this with the latest version of Adrenalin. Should be `amd-software-adrenalin-edition-26.8.1-minimalsetup-260818_web.exe` from https://www.amd.com/en/support/download/drivers.html.

adamlam2-amd:
Which devices are you testing on? I am sure its on many systems, including HaloBox and my personal PC too. 

I think what we can do is add a disclaimer saying 'If this page is not found on your device, you can alternatively change the VGM through BIOS' (although, some BIOS do not have the setting either).

adamlam2-amd:
In addition, HaloBox Linux and non-HaloBox Linux have different paths to change the memory, and thus should maintain the separate images/steps that were used. For HaloBox Linux, we use the 'AMD Ryzen AI Developer Center' that is pre-installed. For non-HaloBox Linux, we use amd-ttm and/or amd-smi (current instructions use amd-ttm, which works for now. Eventually, it will get ported to amd-smi)

PR #941, opened 2026-09-17
restart-runners: force the Linux reboot, matching Windows /f

lucbruni-amd:
`sudo reboot` can hang: a unit that stalls mid-shutdown leaves the machine dark until it is power-cycled. These runners have no hardware watchdog (`/dev/watchdog` absent, no wdt modules), so `RebootWatchdogSec` cannot recover it.

`--force` skips unit teardown, so a hanging unit cannot block the reboot. Processes are still killed and filesystems unmounted or remounted read-only. This is the Linux equivalent of the `/f` the Windows step already uses.

### Verified on a runner

Against a deliberately-planted unit that blocks shutdown for 600s:

| reboot command | result |
| --- | --- |
| `systemctl reboot` (graceful) | blocked the full **600s**, recovered only when the blocker released |
| `systemctl reboot --force` | rebooted in **~23s**, blocker bypassed |

No `plymouth-reboot` involvement and no ext4 recovery or errors on the next boot.

An earlier version of this PR tried a graceful reboot followed by a delayed `--force`. Testing showed that **does not work**: the backgrounded shell is killed at `shutdown.target` before the delay elapses, so the force never runs.


harkgill-amd (approved):
Looks good but is there any way to verify that `sudo --non-interactive systemctl reboot --force` works in the cases that `sudo --non-interactive systemctl reboot` doesn't?

sreeram-11 (approved):

lucbruni-amd:
Changing to force boot instead like Windows with a deliberate blocker succeeded; usurping previous implementation with this one.

PR #964, opened 2026-09-18
restart-runners: detach the Linux reboot so the job can report

lucbruni-amd:
#941 fixed the wedged Linux reboots but broke reporting. `systemctl reboot --force` skips unit teardown, so the runner is killed before it can report the job result: Linux restart jobs report failure even though the reboot itself succeeds. The risk is that a genuine reboot failure is no longer distinguishable.

This keeps the force, so a unit that hangs mid-shutdown still cannot wedge a machine, and adds the property the Windows path already has: the reboot is owned by something outside the job. `systemd-run --on-active=60` creates a transient timer under PID 1, which job cleanup cannot kill. A backgrounded `( sleep 60 ; reboot ) &` stays a child of the step and dies with it.

The 60s delay matches the Windows `/t 60` and must outlast the runner's post-job finalization.

Verified on a Linux runner: the shell session was closed immediately after firing, the machine stayed up the full 60s, rebooted, and the runner came back listening for jobs. Clean shutdown, no filesystem recovery, and the transient unit removed itself.


lucbruni-amd:
Fixes cosmetic reporting issues with `Restart Runners`.

sreeram-11 (approved):

PR #967, opened 2026-09-21
Localize four core playbooks for zh-CN

hazheng1-amd:
Add Simplified Chinese (`zh-CN`) localization for the following core playbooks:

- Generating Images with ComfyUI and Z Image Turbo
- Running and Serving LLMs with LM Studio
- Automating Workflows with n8n and Local LLMs
- Running LLMs on PyTorch with AMD ROCm™ Software

It also localizes the corresponding dependencies and setup guidance for AMD drivers, ComfyUI and its models, LM Studio and Qwen models, Lemonade, Node.js, and Podman.

The content has been adapted for Chinese users, including technical terminology, installation and configuration instructions, model acquisition, and region-appropriate download sources.


localized-playbooks/zh-CN/core/comfyui-image-gen/assets/zimagenode.png
bconsolvo, line 1: Would ask to modify this prompt/image to not be so focused on females (many sample prompts seem to be these days). I would suggest just keeping the landscape beginning from "harbor boats ... etc"

bconsolvo:
@adamlam2-amd Could you review this as well, to the best of your ability?

bconsolvo:
@hazheng1-amd :

A couple things to address before merge:

localized_only isn't set in the 4 playbook.json files. Not a bug — the harness defaults it to true (strict localized), which is the right behavior here. But since these share IDs with the English playbooks, consider setting it explicitly so the intent is obvious to the next editor.

registry.json $schema points to ./registry.schema.json, which doesn't exist in the repo (the English registry has the same dangling ref). Pre-existing — worth adding the file or dropping the ref while we're here.

bconsolvo (changes requested):

